PR #1, opened 2019-08-06
Fix typo in cluster role

2opremio:


hiddeco:
Thanks Fons. I amended the commit instead to keep the history clean, as we are in a preparation phase.

PR #2, opened 2019-08-06
Add manifest-generating API for installation

2opremio:
Similarly to https://github.com/fluxcd/flux/pull/2287 , this PR allows other Go programs to generate the Kubernetes manifests necessary to install the helm operator.

I plan to incorporate this into `fluxctl install --with-helm-op`

PR #4, opened 2019-08-08
 Expand install API and deploy/ options

2opremio:
The PR includes extra content in `deploy/` to illustrate how to enable TLS with Tiller

Also, it adds extra options to the templates in `pkg/install`.

And finally, it fixes the installation API tests. I mistakenly didn't include the fix in #2 (it would help if the tests run for pull requests).


stefanprodan (approved):

pkg/install/templates/tiller-ca-cert-configmap.yaml.tmpl
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: flux-helm-tls-ca-config
+  {{ if .Namespace }}
+  namespace: {{ .Namespace }}{{ end }}
+data:
+  ca.crt: |
+    {{ Base64Encode .TillerTLSCACertContent }}
+{{- end -}}
```
hiddeco: Newline

hiddeco (approved):
Minor comment, other than that looking good.

pkg/install/templates/tiller-ca-cert-configmap.yaml.tmpl
2opremio: It's intentional, so that the file is empty when no TLS is enabled (to discard it). I can change the length check to `<= 1` though

PR #7, opened 2019-08-13
API version bump to `v1` and group domain change

### LINKED ISSUE #9: Promote the HelmRelease CRD to GA
As part of the Flux Helm Operator GA release we should bump the HelmRelease CRD version from `v1beta1` to `v1`. The beta version covers all the Helm options that make sense from a GitOps perspective. We should revisit the CRD when Helm v3 will be available.

hiddeco:
- Change copyright name to "Flux CD contributors"
- Remove `FluxHelmRelease` and the `helm.integratations.flux.weave.works` group domain
- Bump `flux.weave.works/v1beta1` to `helm.fluxcd.io/v1`

Fix: #9


chart/helm-operator/templates/crd.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
   annotations:
     "helm.sh/resource-policy": keep
 spec:
-  group: flux.weave.works
+  group: helm.fluxcd.io
```
stefanprodan, line 15: The version should be `v1` instead of `v1beta1`

deploy/flux-helm-release-crd.yaml
```diff
@@ -2,9 +2,9 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: helmreleases.flux.weave.works
+  name: helmreleases.helm.fluxcd.io
 spec:
-  group: flux.weave.works
+  group: helm.fluxcd.io
```
stefanprodan, line 7: The version should be `v1` instead of `v1beta1`

stefanprodan (changes requested):

stefanprodan (approved):
LGTM

GG Hidde 🥇

PR #29, opened 2019-08-14
Release chart v0.1.0

stefanprodan:
Initial chart release with Helm Operator 1.0.0-rc1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,7 +77,7 @@ jobs:
       - deploy:
           name: Maybe push release image
           command: |
-            if echo "${CIRCLE_TAG}" | grep -Eq "[0-9]+(\.[0-9]+)*(-[a-z0-9]+)?"; then
+            if echo "${CIRCLE_TAG}" | grep -Eq "^[0-9]+(\.[0-9]+)*(-[a-z0-9]+)?$"; then
               echo "$DOCKER_FLUXCD_PASSWORD" | docker login --username "$DOCKER_FLUXCD_USER" --password-stdin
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -77,7 +77,7 @@ jobs:
       - deploy:
           name: Maybe push release image
           command: |
-            if echo "${CIRCLE_TAG}" | grep -Eq "^[0-9]+(\.[0-9]+)*(-[a-z]+)?$"; then
+            if echo "${CIRCLE_TAG}" | grep -Eq "[0-9]+(\.[0-9]+)*(-[a-z0-9]+)?"; then
```
hiddeco: It should be the previous (full match) regex with only `0-9` added, as this will also match e.g. `foo-1.13.0-foo`.

```sh
$ if echo "1.0.0-rc1" | grep -Eq "[0-9]+(\.[0-9]+)*(-[a-z]+)?"; then echo "TRUE"; fi
TRUE

$ if echo "1.0.0-rc1" | grep -Eq "[0-9]+(\.[0-9]+)*(-[a-z0-9]+)?"; then echo "TRUE"; fi
TRUE

$ if echo "1.0.0-rc1" | grep -Eq "^[0-9]+(\.[0-9]+)*(-[a-z]+)?$"; then echo "TRUE"; fi

$ if echo "foo-1.0.0-rc1" | grep -Eq "[0-9]+(\.[0-9]+)*(-[a-z0-9]+)?"; then echo "TRUE"; fi
TRUE
```

hiddeco (approved):

PR #24, opened 2019-08-14
Add descriptions for the CRD fields

stefanprodan:

PR #30, opened 2019-08-15
Chart fixes and improvements 

stefanprodan:
* Fix SSH key mapping
* Restart operator on helm repositories changes
* Add description fields to CRD template
* Add liveness and readiness probes to chart and deploy yaml
* Add `HelmRelease` example to chart notes
* Bump helm chart version


deploy/flux-helm-release-crd.yaml
```diff
@@ -39,57 +39,41 @@ spec:
           required: ['chart']
           properties:
             releaseName:
-              description: The Helm release name. If not supplied, it will be generated by affixing the
```
hiddeco: Why are these removed?

stefanprodan: Looks like make generate-deploy removes them, I have to add the description in the crd template.

stefanprodan: I wander why the CI didn't find this in #24

hiddeco (approved):

PR #31, opened 2019-08-19
fix directory creation for non-root users

knackaron:
When running as non-root, directory permissions have the execute bit stripped.  Maybe something to do with the octal mode wrapping due to the addition operation?  We can just use the octal mode directly to ensure we strip RWX off for other.  Note, the group sticky bit is set below from having `fsGroup` defined on the pod.

Current:
```
sh-4.2$ ls -l /tmp/
total 4
drw-r-S---. 2 1337 1337 4096 Aug 19 01:58 aHR0cH
```

Patched:
```
sh-4.2$ ls -l /tmp/
total 4
drwxr-s---. 2 helm-operator helm-operator 4096 Aug 19 02:02 aHR0cHM6Ly9k
```


stefanprodan (approved):
LGTM 

Thanks @knackaron

PR #33, opened 2019-08-21
Correct a typo in the README

dcherman:


stefanprodan (approved):

stefanprodan:
Thanks @dcherman good catch!

PR #42, opened 2019-09-10
Alpha version of Helm v3 support

hiddeco:
This adds alpha support for Helm v3 by adding a `helmVersion` to
the `HelmRelease` custom resource (valid choices `v2` and `v3`);
for backwards compatibility the default `helmVersion` is `v2`.

Support for both versions is achieved through a `helm.Client`
abstraction interface, with implementations for both versions.

In addition there have been some improvements with regard to
logging metadata, and the Helm v2 client has been upgraded to the
latest version.

Things that still require our attention:

- tests for the client implementations
- extended e2e tests for both Helm versions
- split of the synchronization of charts and helm releases
- finding a way to speed up `hack/` scripts, possibly by giving the deps their own home

Helm v3 install instructions:

```
# install the HelmRelease CRD
kubectl apply -f https://raw.githubusercontent.com/fluxcd/helm-operator/helm-v3/deploy/flux-helm-release-crd.yaml

# install Helm Operator for Helm v3 only
helmv3 upgrade -i helm-operator fluxcd/helm-operator --wait \
--namespace fluxcd \
--set extraEnvs[0].name=HELM_VERSION \
--set extraEnvs[0].value=v3 \
--set image.repository=docker.io/fluxcd/helm-operator-prerelease \
--set image.tag=helm-v3-9908b258
```


pkg/apis/helm.fluxcd.io/v1/types.go
```diff
@@ -208,7 +226,7 @@ func (hr HelmRelease) GetValuesFromSources() []ValuesFromSource {
 }
 
 type HelmReleaseStatus struct {
-	// ReleaseName is the name as either supplied or generated.
+	// GetReleaseName is the name as either supplied or generated.
```
mcristina422, line 229: I think this might have been an accidental change

hiddeco: It was not accidental but changed to get the `HelmRelease` methods in a more uniform shape, as most values are retrieved through `Get<Value>` methods.

stefanprodan (approved):
LGTM 💯

PR #44, opened 2019-09-19
Make update interval a configurable flag

hiddeco:
As in clusters that have many `HelmRelease` resources, the hardcoded
default would put too much pressure on `etcd` due to Tiller collecting
information. By making the interval a configurable flag, people can
tweak the value to find the right balance of (non) pressure and
observability.

Image for testing available as `hiddeco/helm-operator:update-interval-conf-9a48227b`


swade1987 (approved):

stefanprodan (approved):
LGTM

hiddeco:
@stefanprodan given we do not have an 'additional args' value in the Helm chart as we have in the Flux one; do you think adding this to the chart as a dedicated flag is desired, or would you vote for adding an 'additional args' value to the chart instead?

stefanprodan:
I would go with a dedicated flag.

stefanprodan (approved):

PR #47, opened 2019-09-23
Filter out nil release resources during parsing

hiddeco:
As it causes confusion due to a `Object 'Kind' is missing in 'null'` error logged.


stefanprodan (approved):
LGTM

PR #55, opened 2019-09-27
Refactor `hack/` to speed up code generation

hiddeco:


stefanprodan (approved):
Awesome improvement! from 10 minutes to just two 🎉

PR #58, opened 2019-10-02
Update Flux dependency to `v1.15.0`

hiddeco:
This includes import path changes, as Flux moved from the Weaveworks
GitHub organization to the dedicated FluxCD org.


stefanprodan (approved):
LGTM 🚀

PR #61, opened 2019-10-03
Make release build cope with `v` prefixed tags

hiddeco:
As we want to publish our images without the `v` prefix, some
modifications have been made to the `docker/image-tag` helper script to
remove the `v` prefix from the calculated tag.

In addition, the amount of tags it returns has been limited to one, as
multiple tags will make the build fail due to

`-ldflags "-X main.version=<tagA> <tagB>"`

not being a valid argument.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,9 +77,21 @@ jobs:
       - deploy:
           name: Maybe push release image
           command: |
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

PR #65, opened 2019-10-07
Fix git.ssh.secret parameter name in chart documentation

delhage:
The parameter `git.ssh.secretName` is called `git.ssh.secret` in two examples.


hiddeco (approved):
Thank you @delhage :100:

PR #66, opened 2019-10-09
De-experimental-ise `--workers` flag

hiddeco:
This PR both de-experimental-ises the `--workers` flag and adds a FAQ item about tweaking this flag.


stefanprodan (approved):
LGTM

PR #71, opened 2019-10-16
Showing chart name and version in the kubectl get hr

ysaakpr:
I found the `kubectl get hr ` showing the chart and version information at the initial view itself. For usual debugging the chart and the version currently synced from the git would be easier to understand.

### DIFF
--- a/deploy/flux-helm-release-crd.yaml
+++ b/deploy/flux-helm-release-crd.yaml
@@ -19,6 +19,12 @@ spec:
     - JSONPath: .status.releaseStatus
       name: Status
       type: string
+    - JSONPath: .spec.chart.name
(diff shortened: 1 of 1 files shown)


stefanprodan (changes requested):
Can you please move these to `wide` (set `priority: 1`) as these fields are not relevant when using charts from git.

hiddeco (commented):
The change should be made in the template and not directly in the `deploy/` folder, not sure why CI didn't catch this.

https://github.com/fluxcd/helm-operator/blob/master/pkg/install/templates/flux-helm-release-crd.yaml.tmpl

hiddeco:
Pull request has gone stale.

PR #75, opened 2019-10-18
Housekeeping: merge release/1.0.x back into master

stefanprodan:


hiddeco (approved):
Just needs #76 but this will magically appear once merged.

PR #79, opened 2019-10-20
Allow for deploying Tiller as a sidecar.

bmcustodio:
In this PR I propose adding support in the Helm chart for deploying Tiller as a sidecar (restricted to `localhost`). This is inspired by a [comment](https://github.com/fluxcd/flux/issues/1420#issuecomment-426759442) by @justinbarrick, and I believe that, although possibly being an advanced scenario, would be of use for some people ( like me 🙂 ). Tiller is eventually going away, but I believe this may be a useful addition while #8 isn't finished. I've kept this new feature undocumented because it's probably a good idea to "restrict" access to it to whoever finds themselves in need of fiddling with `values.yaml` ( and also because I am not quite sure this will get accepted as a feature 🙂 ).

### DIFF
--- a/chart/helm-operator/templates/deployment.yaml
+++ b/chart/helm-operator/templates/deployment.yaml
@@ -159,6 +159,10 @@ spec:
         {{- else if .Values.allowNamespace }}
         - --allow-namespace={{ .Values.allowNamespace }}
         {{- end }}
+        {{- if .Values.tillerSidecar.enabled }}
+        - --tiller-ip=localhost
+        - --tiller-port=44134
+        {{- else }}
         {{- if .Values.tillerHost }}
         - --tiller-ip={{ .Values.tillerHost }}
         - --tiller-port={{ .Values.tillerPort }}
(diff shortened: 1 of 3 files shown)


bmcustodio:
As a side note, I think this may also be of use to whoever justs wants to try `helm-operator` without having to deploying any other piece of infrastructure (so it might be a good idea to document it, after all).

chart/helm-operator/templates/deployment.yaml
```diff
@@ -177,12 +181,42 @@ spec:
...
+          httpGet:
+            path: /readiness
+            port: 44135
+          initialDelaySeconds: 1
+          timeoutSeconds: 1
+        env:
+        - name: TILLER_NAMESPACE
+          value: {{ .Release.Namespace }}
+        - name: TILLER_HISTORY_MAX
+          value: "0"
```
stefanprodan:
```suggestion
          value: "10"
```
Change the max history setting according to our docs https://docs.fluxcd.io/projects/helm-operator/en/latest/tutorials/get-started.html

chart/helm-operator/values.yaml
```diff
@@ -42,6 +42,16 @@ tls:
...
   hostname: ""
 
+# ADVANCED: Allow for deploying Tiller as a sidecar (restricted to 'localhost' for security reasons).
+# When enabled, either .clusterRole.create should be true or .clusterRole.name should be set to the name of a cluster role granting the required privileges.
+# Please make extra sure you know what you are doing before enabling this. :)
+tillerSidecar:
+  enabled: false
+  image:
+    repository: gcr.io/kubernetes-helm/tiller
+    tag: v2.14.1
```
stefanprodan:
```suggestion
    tag: v2.14.3
```
Match https://github.com/fluxcd/helm-operator/blob/master/docker/helm.version

stefanprodan (changes requested):

bmcustodio:
Thank you @stefanprodan, accepted both changes!

stefanprodan (changes requested):
Please add the `tillerSidecar` fields to the chart readme.

bmcustodio:
@stefanprodan done!

chart/helm-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 appVersion: "1.0.0-rc3"
-version: 0.2.1
+version: 0.3.0
```
stefanprodan: Please undo this change, the version will be bumped when we do a Helm-op release

stefanprodan (changes requested):

stefanprodan (approved):
LGTM

Thanks @bmcstdio

dmccaffery:
If you release this, I might give you a kiss when I meet you next.

```text
O.o
o.O
O.O
>.<
```

Issue #80, opened 2019-10-21
Specify an alternative default branch for HelmRelease objects


nweedo3:
_Preamble: The below is part question, part solution. I've had a dig around the documentation and I can't find any advice or examples on how I would solve this._

---

**Describe the feature**

The Helm Operator should be able to be configured with a default git branch/ref for `HelmRelease` resources specifying the `.spec.chart.git` key.

As `fluxcd/flux` can be configured with a Git branch via `--git-branch`, the operator should also have the option be configured to use a default Git branch/ref other than `master` if the `.spec.chart.ref` field is unspecified.

**What would the new user story look like?**

0. Add an extra command-line and chart flag, something along the lines of `--git-chart-ref-default`. To preserve backwards-compatibility, when the operator is started without this flag, the default value is set to `master` to match the current default [here](https://github.com/fluxcd/helm-operator/blob/203d0df88192fa2bb14860c0f9f99dcac44e8820/pkg/apis/helm.fluxcd.io/v1/types.go#L122).
1. User starts up the operator using the `--git-chart-ref-default` flag, specifying some branch name or reference (i.e. `dev`)
2. A new `HelmRelease` is applied with the `.spec.chart.ref` field unspecified. The operator then retrieves the chart from the default branch specified via the flag (in this case, `dev`).

**Expected behavior**

I'll now go on to explain what problem this solves. Given a typical pipeline of three levels: `dev`, `staging` and `production`:

```
dev --> staging --> production
```

the relevant changes inside the config repository (`charts/` and `releases/`) travel along the pipeline (with each stage represented by a Git branch). To keep the pipeline history clean, `production` should be a subset of the changes in `staging`, and `staging` be a subset of changes of `dev`. To simplify the pipeline, the values provided to each chart are fixed and each pipeline level is represented by one cluster.

The flux instances could be configured in such a way so that `staging` and `production` do not write changes to Git, but mirror a previously verified `dev` configuration:

Pipeline Level | Flux Flags                                     |
-------------- | ---------------------------------------------- |
dev            | `--git-branch=dev`                             |
staging        | `--git-branch=staging --git-readonly=true`     |
production     | `--git-branch=production --git-readonly=true`  |

[...]

PR #83, opened 2019-10-22
Add flag to specify alternative default git ref

nweedo3:
Add the `--git-default-chart-ref` flag which enables the modification
of the ref used to clone a chart if the `.spec.git.ref` field is
unspecified. The original default of `master` is maintained.

Related issue: https://github.com/fluxcd/helm-operator/issues/80

### DIFF
--- a/chart/helm-operator/templates/deployment.yaml
+++ b/chart/helm-operator/templates/deployment.yaml
@@ -145,6 +145,9 @@ spec:
         {{- end }}
         - --git-timeout={{ .Values.git.timeout }}
         - --git-poll-interval={{ .Values.git.pollInterval }}
+        {{- if .Values.git.defaultRef }}
+        - --git-default-ref={{ .Values.git.defaultRef }}
+        {{- end }}
         - --charts-sync-interval={{ .Values.chartsSyncInterval }}
         {{- if .Values.statusUpdateInterval }}
         - --status-update-interval={{ .Values.statusUpdateInterval }}
--- a/cmd/helm-operator/main.go
+++ b/cmd/helm-operator/main.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 
+	"github.com/fluxcd/flux/pkg/checkpoint"
 	fluxhelm "github.com/fluxcd/helm-operator/pkg"
 	"github.com/fluxcd/helm-operator/pkg/chartsync"
 	clientset "github.com/fluxcd/helm-operator/pkg/client/clientset/versioned"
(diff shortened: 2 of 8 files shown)


chart/helm-operator/templates/deployment.yaml
```diff
@@ -145,6 +145,7 @@ spec:
         {{- end }}
         - --git-timeout={{ .Values.git.timeout }}
         - --git-poll-interval={{ .Values.git.pollInterval }}
+        - --git-default-chart-ref={{ .Values.git.defaultChartRef }}
```
stefanprodan: Wrap this in an `if` to maintain backwards compatibility:

```
{{- if .Values.git.defaultChartRef }}
- --git-default-chart-ref={{ .Values.git.defaultChartRef }}
{{- end }}
```

stefanprodan (changes requested):

chart/helm-operator/templates/deployment.yaml
nweedo3: @stefanprodan sure. Would you prefer the entire PR in one commit, or are you fine with me adding the review changes on top of the original commit?

stefanprodan: One commit is fine, amend your commit and force push it. Thanks

pkg/apis/helm.fluxcd.io/v1/types.go
```diff
@@ -5,7 +5,7 @@ import (
 	"strings"
 
 	"github.com/ghodss/yaml"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
stefanprodan: Please remove this alias, it has no effect since it's the same with the package name.

stefanprodan (changes requested):

nweedo3:
@stefanprodan the changes should be good to go now - thanks for the review.

nweedo3:
@stefanprodan do you have any further updates on this? Thanks.

stefanprodan (approved):
LGTM

cmd/helm-operator/main.go
```diff
@@ -112,6 +113,7 @@ func init() {
 
 	gitTimeout = fs.Duration("git-timeout", 20*time.Second, "duration after which git operations time out")
 	gitPollInterval = fs.Duration("git-poll-interval", 5*time.Minute, "period on which to poll git chart sources for changes")
+	gitDefaultChartRef = fs.String("git-default-chart-ref", "master", "ref to clone chart from if ref is unspecified in a HelmRelease")
```
hiddeco: I think the length of the flag can be reduced to `--git-default-ref`, as the only Git thing the operator does is syncing charts.

hiddeco (approved):
Thank you @niall-weedon :1st_place_medal:

nweedo3:
No problem @hiddeco, thanks for the review! :)

PR #84, opened 2019-10-24
Add release count metric

larsroger:
As we want to know the number of relases performed by the Helm
Operator, a counter metric (`releases_performed_total`) was added
to keep track of this number.
Both installations and upgrades increase the counter.

**Examples of how to use the metric in a PromQL query:**
Using the metric "as is"
(`flux_helm_operator_releases_performed_total`) returns the total
number of releases since the Helm Operator was started.

To get the total number of releases performed during the last
7 days:
`sum(max_over_time(flux_helm_operator_releases_performed_total[7d]))`

Co-authored-by: larsroger <larsroger@users.noreply.github.com>

### DIFF
--- a/pkg/release/metrics.go
+++ b/pkg/release/metrics.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/go-kit/kit/metrics/prometheus"
 	stdprometheus "github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/release/metrics.go
```diff
@@ -36,3 +46,7 @@ func ObserveRelease(start time.Time, action Action, dryRun, success bool, namesp
 		LabelReleaseName, releaseName,
 	).Observe(time.Since(start).Seconds())
 }
+
+func RecordRelease() {
+	go releasesPerformed.Inc()
```
cep21, line 51: It should be quicker to increment an integer than spinning up a goroutine.  Why do you do this inside `go` ?

PR #92, opened 2019-10-30
Fix upgrade command for install instructions

joazvsoares:


hiddeco (approved):
Thank you @joazvsoares :medal_sports:

PR #99, opened 2019-11-08
Decouple release reconciliation from chart source sync

### LINKED ISSUE #98: Decouple release reconciliation from chart source sync
In the current state of the code base the `chartsync` package tries to take care of too many things to make it easy to work with, as it:

1. keeps track of chart source changes in Git
2. provides the tools to fetch chart sources from HTTP(S) Helm chart registries
3. re-conciliates `HelmRelease` resources to match the expected Helm release state

As 3 has much overlap with the `release` package, and for Helm v3 support a client abstraction was introduced, moving this logic into the `release` package would lead to a severe improvement in code readability, and reduce the overall complexity.

The end result would make the context of the `chartsync` package _chart source sync_ only, while the `release` package takes care of releasing the `HelmRelease` resource with Helm (==syncing) using the synced chart.

hiddeco:
This PR decouples the synchronization of charts from git sources from the synchronization of `HelmRelease` resources to Helm releases and/or cluster resources.

The synchronization of charts from Git sources has been rewritten so that we only do bookkeeping of git references for `HelmRelease` resources, but no longer maintain a working directory, this copy is instead made on request based on the references we have in our books at that moment.

The release synchronization is moved to the `release` package, and has been rewritten so that implementers no longer have to decide what actions they want to perform, they can simply call `Sync` (install or upgrade on creation and/or update of `HelmRelease` resource) or `Uninstall` (uninstall on delete of `HelmRelease` resource).

Fixes #98


pkg/chartsync/download.go
```diff
@@ -49,6 +35,20 @@ func ensureChartFetched(base string, source *helmfluxv1.RepoChartSource) (string
...
 
+// makeChartPath gives the expected filesystem location for a chart,
+// without testing whether the file exists or not.
+func makeChartPath(base string, source *helmfluxv1.RepoChartSource) string {
+	// We don't need to obscure the location of the charts in the
+	// filesystem; but we do need a stable, filesystem-friendly path
+	// to them that is based on the URL.
+	repoPath := filepath.Join(base, base64.URLEncoding.EncodeToString([]byte(source.CleanRepoURL())))
+	if err := os.MkdirAll(repoPath, 00750); err != nil {
+		panic(err)
```
2opremio: I know you just moved this but I am generally against panics. It's OK to exit, but with a friendlier error than a stack trace

pkg/chartsync/git.go
```diff
@@ -0,0 +1,407 @@
...
+		// We did not find a mirror; request one, return, and wait for
+		// signal.
+		c.maybeMirror(mirror, hr.Spec.GitURL)
+		return nil, false
+	}
+
+	if s, _ := c.syncGitChartSource(repo, hr); s != nil {
+		return s, true
+	}
+	return nil, false
```
2opremio: You can replace all this by:
```go
s, _ := c.syncGitChartSource(repo, hr)
return s, s != nil 
```

pkg/chartsync/git.go
```diff
@@ -0,0 +1,407 @@
...
+	}()
+}
+
+// Load returns a pointer to the requested `GitChartSource` for the given
+// `v1.HelmRelease` if found, and a boolean indicating success.
+func (c *GitChartSourceSync) Load(hr *v1.HelmRelease) (*GitChartSource, bool) {
+
+	// Check if we have a source in store and return if it still equals
+	// to what is configured in the source.
+	c.sourcesMu.Lock()
```
2opremio: You may want to consider using a `RWMutex` and `RLock()`ing here instead of `Lock()`ing

pkg/chartsync/git.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+}
+
+// Run starts the sync of `GitChartSource`s.
+func (c *GitChartSourceSync) Run(stopCh <-chan struct{}, errCh chan error, wg *sync.WaitGroup) {
+	c.logger.Log("info", "starting sync of git chart sources")
+
+	wg.Add(1)
+	go func() {
+		defer runtime.HandleCrash()
```
2opremio: Is this really necessary?

pkg/chartsync/git.go
```diff
@@ -0,0 +1,407 @@
...
+		defer func() {
+			c.mirrors.StopAllAndWait()
+			wg.Done()
+		}()
+
+		for {
+			select {
+			case changed := <-c.mirrors.Changes():
+				for mirrorName := range changed {
+					mirror, ok := c.mirrors.Get(mirrorName)
```
2opremio: This function is pretty large, I would extract the body of the for loop into a separate method (e.g. `processChangedMirror()`)

2opremio: Maybe it's better to only factor out the `ok == true` part of the body.

pkg/chartsync/git.go
hiddeco: I do not think so, I think it is there for historical reasons, as the `panic` you commented on earlier in a previous version of the operator could have been triggered by this loop.

pkg/chartsync/git.go
```diff
@@ -0,0 +1,407 @@
...
+		return nil, false
+	}
+
+	// Acquire sources lock and attempt to find a source in store.
+	c.sourcesMu.Lock()
+	s, ok := c.sources[hr.ResourceID().String()]
+	if !ok {
+		// No source found in store;
+		// create the boilerplate of a new one.
+		s = &GitChartSource{Mutex: sync.Mutex{}, Mirror: mirrorName, Remote: source.GitURL}
```
2opremio: I don't think it's necessary to initialize the mutex

pkg/chartsync/git.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+	c.logger.Log("info", "finished syncing git chart sources")
+}
+
+// syncGitChartSource synchronizes the source in store with the latest
+// HEAD for the configured ref in the given repository. But only if it
+// has seen commits for paths we are interested in. It returns the
+// synchronized source or nil and a boolean indicating if the source
+// was updated.
+func (c *GitChartSourceSync) syncGitChartSource(repo *git.Repo, hr *v1.HelmRelease) (*GitChartSource, bool) {
```
2opremio: This function is a bit too long. Anything you can factor out from here?

pkg/chartsync/git.go
```diff
@@ -0,0 +1,407 @@
...
+		// Failed creating a git clone at the given ref HEAD.
+		return s, false
+	}
+
+	if oldExport := s.Export; oldExport != nil {
+		// Defer clean-up of old export.
+		defer oldExport.Clean()
+	}
+
+	defer func() {
```
2opremio: Any reason for not simply doing this at the end of the function instead of using `defer`?

pkg/chartsync/git.go
```diff
@@ -0,0 +1,407 @@
...
+	c.sourcesMu.Lock()
+	s, ok := c.sources[hr.ResourceID().String()]
+	if !ok {
+		// No source found in store;
+		// create the boilerplate of a new one.
+		s = &GitChartSource{Mutex: sync.Mutex{}, Mirror: mirrorName, Remote: source.GitURL}
+		c.sources[hr.ResourceID().String()] = s
+	}
+	// Acquire source lock and unlock sources lock.
+	s.Lock()
```
2opremio: Locking the entry while holding map lock means that doing the opposite (locking the map while holding the lock of an entry) can cause a deadlock. I haven't been able to see that happening in the code, but it's something to keep in mind.

pkg/chartsync/git.go
hiddeco: I want to set the condition on the `HelmRelease` _after_ the source has become available so that the timestamp of the condition update is technically correct. Which isn't the case as long as this method holds the lock.

pkg/release/release.go
```diff
@@ -1,556 +1,342 @@
...
-		helm.StatusPendingRollback:
-		return nil, fmt.Errorf("operation pending for release (%s)", rel.Info.Status)
-	default:
-		return nil, fmt.Errorf("current state prevents it from being upgraded (%s)", rel.Info.Status)
-	}
-}
+		// Lock the resource so that it does not get updated by
+		// the chart sync process.
+		s.Lock()
+		defer s.Unlock()
```
2opremio: Do you need to hold the lock throughout the execution of the rest of function?

hiddeco: Yes, as otherwise a signal could replace the working clone and clean up the old one while the release has not happened yet. Causing the path to the chart to no longer exist.

pkg/chartsync/git.go
```diff
@@ -0,0 +1,407 @@
...
+// GitConfig holds the configuration for git operations.
+type GitConfig struct {
+	GitTimeout      time.Duration
+	GitPollInterval time.Duration
+}
+
+// GitChartSource holds the source references to a Helm chart in git,
+// and a minimal working clone.
+type GitChartSource struct {
+	sync.Mutex
```
2opremio: Exposing the mutex externally makes it hard to reason about concurrency issues. Also it's not obvious (to me at least) when should one hold the lock (at least without looking at the implementation of `GitChartSourceSync`). During any action on Export? 

Maybe we can provide the same concurrency guarantees through a more intuitive/encapsulated abstraction. For instance, is it possible to provide the same guarantees by providing methods for `GitChartSource` instead?

pkg/chartsync/git.go
2opremio: We are probably talking about a sub-millisecond time difference, is that really worth it? At least I think we should add a comment.

pkg/chartsync/git.go
hiddeco: One should hold the lock for the duration one works with the source, as the `Export` has a relationship with all other fields in the structure. Providing multiple methods is thus not an option.

The only way I can think of encapsulating this is to create a callback method, which acquires the lock and defers the unlock. I may however be short sighted in options/possibilities.

pkg/chartsync/git.go
```diff
@@ -0,0 +1,337 @@
...
+	s.head = head
+	c.store(hr, s)
+	return s, changed, nil
+}
+
+// maybeMirror requests a new mirror for the given remote. The return value
+// indicates whether the repo was already present (`true` if so,
+// `false` otherwise).
+func (c *GitChartSync) maybeMirror(mirrorName string, remote string) bool {
+	ok := c.mirrors.Mirror(
```
2opremio, line 284: Maybe this isn't a new problem but, I only see a `StopOne()` counterpart to this `Mirror()` call in `Delete()` which seems to only be invoked when a `HelmRelease()` is deleted.

Doesn't this mean that there will be a mirror leak if the `git` entry of a `HelmRelease` is changed (or removed)?

pkg/chartsync/git.go
```diff
@@ -0,0 +1,337 @@
...
+// It returns a boolean indicating a successful removal (`true` if so,
+// `false` otherwise).
+func (c *GitChartSync) Delete(hr *v1.HelmRelease) bool {
+	c.sourcesMu.Lock()
+	defer c.sourcesMu.Unlock()
+
+	// Attempt to get the source from store.
+	source, ok := c.sources[hr.ResourceID().String()]
+	if ok {
+		// Remove the in store source.
```
2opremio: Remove the ... release?

pkg/chartsync/git.go
2opremio: Additionally, `Delete()` doesn't help once the `git` field is modified, since it uses `helmReleasesForMirror()` which relies in live data, and not the mirrors created by the `HelmRelease()`

2opremio: If we don't want to track changes on the`git` field explicitly, I think this can be fixed by periodically comparing the output of `helmReleasesForMirror()` on `c.mirrors`. I don't think it's ideal though.

pkg/chartsync/git.go
```diff
@@ -0,0 +1,337 @@
...
+// when the update is relevant to the release.
+func (c *GitChartSync) Run(stopCh <-chan struct{}, errCh chan error, wg *sync.WaitGroup) {
+	c.logger.Log("info", "starting sync of git chart sources")
+
+	wg.Add(1)
+	go func() {
+		defer func() {
+			c.mirrors.StopAllAndWait()
+			wg.Done()
+		}()
```
2opremio: Why not place this code directly in the `case <-stopCh:` instead of using a deferred function call?

pkg/chartsync/git.go
```diff
@@ -0,0 +1,337 @@
...
+			// stop and delete the mirror.
+			c.mirrors.StopOne(source.mirror)
+		}
+	}
+	return ok
+}
+
+// SyncMirrors instructs all git mirrors to sync from their respective
+// upstreams.
+func (c *GitChartSync) SyncMirrors() {
```
2opremio, line 197: This seems to be needed because of https://github.com/fluxcd/helm-operator/commit/c20e5c84c2e4580c3ab5eba157250c012a1ae9c3 ,  but I cannot find a matching PR. Would you mind clarifying why the http endpoint was added?

pkg/chartsync/git.go
```diff
@@ -0,0 +1,337 @@
...
+
+// GitConfig holds the configuration for git operations.
+type GitConfig struct {
+	GitTimeout      time.Duration
+	GitPollInterval time.Duration
+}
+
+// GitChartDir holds a `git.Export` of a chart repository,
+// and the commit hash of the HEAD of this export.
+type GitChartDir struct {
```
2opremio: I was going to ask for a name change (something like `GitChartWorkingDir`) but, after seeing its use I think its overkill to have an explicit structure. I would recommend getting rid of it and returning an `Export` and the `Head` explicitly (`ChartPath` can be implemented by the consumer).

pkg/chartsync/git.go
```diff
@@ -0,0 +1,337 @@
...
+				c.logger.Log("info", "stopping sync of git chart sources")
+				return
+			}
+		}
+	}()
+}
+
+// ChartDir returns a newly exported `GitChartDir` for the given `v1.HelmRelease`
+// at the latest HEAD, or an error.
+func (c *GitChartSync) ChartDir(hr *v1.HelmRelease) (GitChartDir, error) {
```
2opremio: I would rename this to something with a verb (since it requires some work) and more descriptive. Something like `GetChartSnapshotDir()` or `GetMirrorCopy()` or `TakeChartMirrorSnapshot()` or `TakeChartSnapshot()`

pkg/chartsync/git.go
```diff
@@ -0,0 +1,337 @@
...
+func (c *GitChartSync) get(hr *v1.HelmRelease) (sourceRef, bool) {
+	c.sourcesMu.RLock()
+	defer c.sourcesMu.RUnlock()
+	if s, ok := c.sources[hr.ResourceID().String()]; ok && s.forHelmRelease(hr) {
+		return s, ok
+	}
+	return sourceRef{}, false
+}
+
+func (c *GitChartSync) store(hr *v1.HelmRelease, s sourceRef) {
```
2opremio: This is only used once, so I would simply inline it.

pkg/chartsync/git.go
```diff
@@ -0,0 +1,337 @@
...
+			if err != nil {
+				continue // this should never happen
+			}
+			// Schedule release sync by adding it to the queue.
+			c.releaseQueue.AddRateLimited(cacheKey)
+		}
+	}
+}
+
+func (c *GitChartSync) get(hr *v1.HelmRelease) (sourceRef, bool) {
```
2opremio: This is only used once, so I would simply inline it.

pkg/chartsync/git.go
```diff
@@ -0,0 +1,337 @@
...
+type GitChartSync struct {
+	logger log.Logger
+	config GitConfig
+
+	lister lister.HelmReleaseLister
+
+	mirrors *git.Mirrors
+
+	sourcesMu sync.RWMutex
+	sources   map[string]sourceRef
```
2opremio: I think it's worth adding a comment about the map being indexed by the `ResourceID` of `HelmReleases` and probably also renaming it to `releaseSources` or `releaseSourcesByID`

pkg/chartsync/git.go
2opremio: or at least rename it to `getSource`

pkg/chartsync/git.go
2opremio: or at least rename it to `storeSource`

pkg/chartsync/git.go
2opremio: Now I see that `forHelmRelease()` is used in `get()` to take into account that  the repo of `HelmRelease`s  can change with respect to what's book-kept in `GitChartSync`, but I don't think we do anything to correct divergences.

pkg/chartsync/git.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+// sync synchronizes the record we have for the given `v1.HelmRelease`
+// with the given mirror. It always updates the HEAD record in the
+// `sourceRef`, but only returns `true` if the update was relevant for
+// the release (e.g. a change in git the chart source path, or a new
+// record). In case of failure it returns an error.
+func (c *GitChartSync) sync(hr *v1.HelmRelease, mirrorName string, repo *git.Repo) (sourceRef, bool, error) {
+	source := hr.Spec.GitChartSource
+	if source == nil {
```
2opremio, line 228: It's strange that the method needs to check for the `GitChartSource` and also provide a `mirrorName`, which should be a subproduct of the former).

2opremio: I would change the interface to `sync(hrID resource.ID, mirrorName string, repo *git.Repo) ` 

But, that's not currently possible, because `sync` also consistency-checks that the mirror we have for the Helm release (in `get()`s invocation to `forHelmRelease()`).

However, I don't think that consistency-checking is the direct responsibility of `sync()` (or at least it makes it harder to understand what's going on since it's buried when invoking `get()`).

I think that we should take car of divergences explicitly, while also garbage collecting orphan mirrors (i.e. https://github.com/fluxcd/helm-operator/pull/99/files#r347533178 )

pkg/chartsync/git.go
```diff
@@ -0,0 +1,337 @@
...
+	if !ok {
+		c.logger.Log("info", "started mirroring new remote", "remote", remote, "mirror", mirrorName)
+	}
+	return ok
+}
+
+// helmReleasesForMirror returns a slice of `HelmRelease`s that make
+// use of the given mirror.
+func (c *GitChartSync) helmReleasesForMirror(mirror string) ([]*v1.HelmRelease, error) {
+	hrs, err := c.lister.List(labels.Everything())
```
2opremio, line 296: Isn't it a bit overkill to list all the helm releases every time? I don't know the operator well, but I presume we use a `HelmRelease` informer, which we can use to cache the mirrors and keep them updated.

2opremio: We could use the same cache to make sure that `c.sources` are up to date, allowing to garbage collect mirrors and (maybe) allowing to remove the consistency check in `get()`

pkg/chartsync/git.go
hiddeco: It is the operator’s equivalent of `fluxctl sync`, but via an HTTP endpoint, to give users the option to instantly trigger the refresh from git mirrors from e.g. CI, instead of waiting for the poll interval to kick in.

2opremio:
I think that the implementation has been drastically improved and looks much cleaner 🏅 , particularly when it comes to concurrency

However, this has allowed me to focus on other parts, particularly the consistency of `git references from`HelmReleases` which I think can be improved.

pkg/chartsync/git.go
hiddeco: Original PR: https://github.com/fluxcd/flux/pull/1776

pkg/chartsync/git.go
hiddeco: One of the few original lines from the original code.

Historically looking back, I think this is because the previous `panic` possibility, which would still make the `defer` happen (but would not reach the `stopCh`).

pkg/chartsync/git.go
hiddeco: > Maybe this isn't a new problem but, I only see a `StopOne()` counterpart to this `Mirror()` call in `Delete()` which seems to only be invoked when a `HelmRelease()` is deleted.

The current `master` version does not stop any mirrors, they continue to live forever. I will let your comments simmer in my head for a while.

pkg/chartsync/git.go
hiddeco: This is a `HelmReleaseLister`, which is nothing more then a named `cache.Indexer`, we basically loop over in-memory `HelmRelease` resources at close-to-zero cost.

pkg/chartsync/git.go
hiddeco: I felt that changing the name to `releaseSourcesByID` made the comment itself obsolete.

pkg/chartsync/git.go
hiddeco: We are not deleting the release here but our indexation / bookkeeping of the git source. I think the comment does not add much now that the code has been reduced.

pkg/chartsync/git.go
hiddeco: (I can also give you some historical insights and tell you that before my previous refactor round, we used the client for this type of operations 🤫).

pkg/chartsync/git.go
2opremio: Ah, you meant the _in-store_ source (sorry, I should had understood)

pkg/chartsync/git.go
2opremio: Of course, I forgot this repo was split out recently, sorry.

pkg/chartsync/git.go
hiddeco: Yes, I already thought the missing hyphen was the actual issue.

pkg/chartsync/git.go
2opremio: > we basically loop over in-memory HelmRelease resources at close-to-zero cost.

Ah, I see, I thought it involved calling the client 👍

pkg/chartsync/git.go
hiddeco: There is another reason this is not possible, without the `v1.HelmRelease` or a copy of the `v1.GitChartSource`, it is not possible to detect changes as done on L286, nor is it possible to construct a new `sourceRef` as done on L265.

hiddeco:
@fons I would like to propose to merge this _as is_, and tweak the bits around garbage collection and/or the imperfect clean up of mirrors at a later stage. This PR is holding back the developments around Helm v3 - which I would like to focus on the rest of this week, and the current state of the PR is arguably already ten times better as what it was.

We can work out the missing piece of the puzzle in an issue, as we both do not seem to have a direct answer to the question what the best way forward is for this given the triangular relationship of `HelmRelease` -> `sourceRef` <- `mirror(Signal)`. I also think having some time off from this code, while focusing on other elements of the operator, will probably lead to a solution faster than being stuck on minor details.

2opremio:
Sounds good! Can you create a ticket for it?

2opremio (approved):

PR #104, opened 2019-11-13
pkg: only use official flux

carnott-snap:
It appears in a previous refactor, after the repo name migration, that
one remaining reference to github.com/weaveworks/flux was left in
source. This change fixes that, and removes the dependency from the
go.mod.


hiddeco (approved):
Thank you @carnott-snap :tulip:

Issue #103, opened 2019-11-12
End-to-end tests


hiddeco:
Given that there has been much progress on the end-to-end testing side of Flux, and we can use a lot of the libraries there to scaffold the tests here, the easiest way to get a validation of all (new) features is to test them in an end-to-end setting.

The minimum tests required to get an overview are:

- Installation
  - [x] Helm v2 git chart installation
  - [x] Helm v2 repo chart installation
  - [x] Helm v3 git chart installation
  - [x] Helm v3 repo chart installation
- Upgrade
  - [x] Helm v2 upgrade
  - [x] Helm v2 upgrade triggered by git chart source change
  - [x] Helm v3 upgrade
  - [x] Helm v3 upgrade triggered by git chart source change
- Rollback
  - [x] Helm v2 rollback
  - [x] Helm v3 rollback
  - [x] Rollback halts sequential releases
  - [x] Change to chart and/or values continues release
- Uninstall
  - [ ] Helm v2 uninstall
  - [ ] Helm v3 uninstall
- Dependencies
  - [ ] Helm v2 git chart source dependencies are updated
  - [ ] Helm v3 git chart source dependencies are updated
- Other
  - [ ] Default Helm version setting
  - [ ] HTTP sync request
  - [ ] ...

PR #110, opened 2019-11-20
End-to-end test framework scaffold

hiddeco:
Scaffolding for #103.

This lacks any tooling around working with multiple Helm versions.


docker/kubectl.version
```diff
@@ -1,5 +1,4 @@
-KUBECTL_VERSION=v1.13.8
-KUBECTL_CHECKSUM_amd64=b24d2f026064945a7e1571fd413bd974e165ca4514e35a3dae7df8cbf97bd283169c40e7720cbf40d6fdce2b37483df8a02a58c3a9cc9fc08aa386b43a467c0a
-KUBECTL_CHECKSUM_arm=dc60bdf00e6c7806e3c11e4f73e2ff27a603e968f22567d8c87ed5ece04263e557cb0df8d7b5196bc28a96b0b1f8ae104c503e9b9d90ce77c25e85954b54e178
-KUBECTL_CHECKSUM_arm64=00a98acd51107d1cb935cfc07ca31487290412f92ac34a91ec8c7f4b802bf798a7cc9cac22978a92001641c3c99a651c7f9a9f7f64f7cc4839a9d26021667a3b
-
+KUBECTL_VERSION=v1.14.9
```
2opremio, line 1: Is the version bump intentional?

test/e2e/10_helm_chart.bats
```diff
@@ -0,0 +1,36 @@
...
+  install_tiller
+  install_helm_operator_with_helm
+  kubectl create namespace "$DEMO_NAMESPACE"
+}
+
+@test "Helm chart installation smoke test" {
+  # The gitconfig secret must exist and have the right value
+  poll_until_equals "gitconfig secret" "${GITCONFIG}" "kubectl get secrets -n ${E2E_NAMESPACE} gitconfig -ojsonpath={..data.gitconfig} | base64 --decode"
+
+  poll_until_true "namespace $DEMO_NAMESPACE" "kubectl describe ns/$DEMO_NAMESPACE"
```
2opremio: Isn't this testing tiller instead of the helm operator? I would instead  check with the helm client that the release is made

2opremio: Actually, the releases are applied below, so I would simply remove this

test/e2e/10_helm_chart.bats
```diff
@@ -0,0 +1,36 @@
...
+
+  # Apply the HelmRelease fixtures
+  kubectl apply -f "$FIXTURES_DIR/releases" >&3
+
+  poll_until_true 'podinfo-helm-repository HelmRelease' 'kubectl -n demo get helmrelease/podinfo-helm-repository | grep -e "deployed"'
+  poll_until_true 'podinfo-git HelmRelease' 'kubectl -n demo describe helmrelease/podinfo-git | grep -e "deployed"'
```
2opremio: Doesn't this simply test that the releases were applied? Shouldn't we test that the operator applied them instead?

2opremio:
I think we should share the scaffolding with Flux instead of (mostly) duplicating the files.

How about using a common repository and downloading an exported tarball (or even cloning the repo) locally before running the tests? It's a poor man's external library import system, but I think it's better than simply duplicating the files

docker/kubectl.version
hiddeco: Yes, it even has a dedicated commit (`5cccc11`). The previous version did not support `kubectl apply -k`.

test/e2e/10_helm_chart.bats
hiddeco: No, it tests that the releases have been made by the operator and were successful, hence the `| grep -e "deployed"`.

hiddeco:
@2opremio I think for the small amount of overlap (as only `lib/poll.bash` and `lib/template.bash` are 1:1 copies), having duplicates is easier to work with and more flexible than a common repository.

test/e2e/10_helm_chart.bats
2opremio: Ah, sorry, I missed that. It would be cleaner to use `poll_until_equals` though and extracting the status field.

hiddeco: I forgot about the existence of that method while modifying what I copied from Flux. Will use that instead, much better indeed.

2opremio:
Uhm, the basic structure of the libraries is still the same, and maybe it could be adapted to have a lower divergence (sharing more code). But I will let you decide :)

2opremio (approved):
LGTM. Only a couple of optional change suggestions.

hiddeco:
@2opremio agreed, but to be able to adapt it to have a lower divergence, one first needs to know what they have in common or how they utilize things, which is in my opinion something that will not be clear until both projects have a decent amount of tests.

I can create a follow up issue for #103 (probably on the Flux side of issue tracking) to look into this once the operator has a decent basic test coverage.

PR #111, opened 2019-11-21
Proper support for enabling Helm client versions

hiddeco:
This commit adds proper support for enabling Helm client versions
and configuring the default Helm client version.

The enabled versions can be set by either setting the
`--enabled-helm-versions` flag, or the `HELM_VERSION` environment
variable.

The first version given will become the default client for
`HelmRelease` resources that do not have a `helmVersion` declared in
their spec.

### DIFF
--- a/cmd/helm-operator/main.go
+++ b/cmd/helm-operator/main.go
@@ -66,6 +66,7 @@ var (
 	listenAddr *string
 
 	enabledHelmVersions *[]string
+	defaultHelmVersion  *string
 )
 
 const (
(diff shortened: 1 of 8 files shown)


cmd/helm-operator/main.go
```diff
@@ -129,10 +130,11 @@ func main() {
 		os.Exit(0)
 	}
 
-	// TODO(hidde): hacked in so that we can control the Helm version with an
-	// environment variable so that we do not have to make any chart modifications.
-	if ver := getEnvAsSlice("HELM_VERSION", []string{}); len(ver) == 1 {
-		enabledHelmVersions = &ver
+	// Support enabling the Helm supported versions through an
+	// environment variable.
+	helmVersionEnv := getEnvAsSlice("HELM_VERSION", []string{})
```
2opremio, line 135: Not part of the PR, but wouldn't it be simpler to only use a flag?

Why should there be two ways of doing it?

2opremio: Also, it raises the question of what should take precedence, the flag or the env variable.

hiddeco: A flag always overwrites a set environment variable, that is why there is an explicit check to determine if the default flag value has been changed (`fs.Changed`).

The environment variable was introduced to support setting the value without making any additional adjustments to our Helm chart, and is now being used in some of Stefan his tutorials to only target Helm version 3, which means I can not remove it at the moment.

2opremio (approved):

Issue #69, opened 2019-10-11
Helm release revision created every minute if secretKeyRef used


brew:
If a `secretKeyRef` is defined in the HelmRelease manifest, the helm-operator creates a new helm release every minute.

#### with `secretKeyRef`

With the following HelmRelease file pushed to my repo:

```yaml
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  name: podinfo
  namespace: stage
  annotations:
    fluxcd.io/ignore: "false"
spec:
  releaseName: podinfo-stage
  targetNamespace: stage
  chart:
    git: git@github.com:myorg/infrastructure
    ref: 14/gitops
    path: k8s/charts/podinfo
  valuesFrom:
  - configMapKeyRef:
      name: podinfo-stage-values
      key: stage.values.yaml
  - secretKeyRef:
      name: podinfo-stage-secrets
      key: stage.secrets.yaml      
```

The helmrelease is created, and podinfo workloads created as expected, including a helm release secret `secret/podinfo-stage.v1`. After a minute another helm release secret is created, `secret/podinfo-stage.v2`. This carries on every minute.

```sh
NAME                           TYPE                                  DATA   AGE
secret/default-token-whxz6     kubernetes.io/service-account-token   3      27h
secret/podinfo-stage-secrets   Opaque                                1      19m
secret/podinfo-stage.v1        helm.sh/release                       1      2m23s
secret/podinfo-stage.v2        helm.sh/release                       1      91s
secret/podinfo-stage.v3        helm.sh/release                       1      30s
```

#### without `secretKeyRef`

Clean up the release by setting `fluxcd.io/ignore: "true"`, commit and push, then use `kubectl` to delete the helmrelease.

Edit the HelmRelease file to remove the `secretKeyRef` section (reset `fluxcd.io/ignore: "false"`), and commit and push it.

This creates the helmrelease again, and the podinfo workloads. This time, only one helm release secret is created, and no other releases are created after a minute.

**Additional context**
Add any other context about the problem here, e.g 
- Helm operator version: chart from commit 8f5bbeb9ec47569280df94cb0b3470611f826527 with the following values:

```yaml
git:
  ssh:
    secretName: flux-git-deploy
  pollInterval: 1m

chartsSyncInterval: 1m

configureRepositories:
  enable: true
  repositories:
    - name: https://charts.bitnami.com/bitnami
      url: https://charts.bitnami.com/bitnami
[...]

PR #117, opened 2019-11-24
helm/v3: always YAML parse returned release values

hiddeco:
This commit adds a `configToGenericValues` method to the v3 client
layer that forces the returned release values to be (re)parsed as
YAML so that the value types of the nested map always equal to what
they would be when returned from storage.

The main reason we have to do this is because a dry-run release seems
to skip this step and return the values directly from memory, causing
difficulties during the comparison we make to see if we actually have
to make a release, as e.g. float values will be strings when returned
from storage (current release), but in the values of the dry-run
release they will still be floats.

This PR likely also addresses #69, but this issue should not be closed
before we have confirmed this, and has been covered by tests.

Image for testing available: `hiddeco/helm-operator:helm-v3-v3-value-types-334af866`

### DIFF
--- a/pkg/helm/v3/release.go
+++ b/pkg/helm/v3/release.go
@@ -2,6 +2,7 @@ package v3
 
 import (
 	"helm.sh/helm/pkg/chart"
+	"helm.sh/helm/pkg/chartutil"
 	"helm.sh/helm/pkg/release"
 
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

PR #118, opened 2019-11-24
Build: include both major Helm version binaries in image

hiddeco:
This PR adds the Helm `v2` _and_ `v3` binaries to the Docker image. It also adds the base for running the end-to-end tests targeting a specific Helm version, but does not yet take the targeted version into account during the installation procedure itself.


stefanprodan:
The e2e tests expects a binary called `helm`, can you rename it to `helm2` so we get the tests fixed please?

stefanprodan:
@hiddeco will the helm deps build still work after the renaming?

stefanprodan (approved):
LGTM

PR #119, opened 2019-11-24
helm/v2: mimic v3 `--atomic` behaviour in client

hiddeco:
This commit mimics the Helm v3 `--atomic` behaviour in the `v2` client layer, so that we can rely on this behaviour in the `release` package and do not have to take care of the removal of a release ourselves when something fails.


stefanprodan (approved):
LGTM

PR #125, opened 2019-11-27
Added extraVolumes and extraVolumeMounts for Helm Operator

binboum:
Hi,

It's possible to mount volumes on the Flux daemon but not on Helm Operator.
In my context, I proxify the GIT protocol, and I need to mount a configuration file.


stefanprodan (approved):
LGTM

Issue #130, opened 2019-12-03
Existing resource conflict with sealed-secrets release


stefansedich:
**Describe the bug**
Getting the following with a new sealed-secrets release, it initially instally fine but then on the next loop will spit out the following error in the logs:

```
ts=2019-12-03T06:28:01.289963574Z caller=release.go:182 component=release release=sealed-secrets-controller targetNamespace=system resource=system:helmrelease/sealed-secrets-controller helmVersion=v3 error="failed to determine if the release should be synced" err="failed to upgrade chart for release [f97b6911-155a-11ea-96fd-02d8765e21f2]: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: kind: ClusterRole, namespace: , name: secrets-unsealer"
```

In this case it is right after the initial install and there are actually no changes, from what I can tell if I make an actual change to the helm-release (say I bump the version down) it will actually re-deploy without issue.

**To Reproduce**
Steps to reproduce the behaviour:
1. Deploy the following release https://gist.github.com/stefansedich/2e833d1cd26315169563554cd30f4751
2. Witness it installs fine
3. Wait for the next helm-operator loop to run
4. Witness the error in helm-operator logs
5. Wait a few more loops
6. Witness more errors
7. Change helmrelease image tag to v0.9.5
8. Witness it deploys fine

**Expected behavior**
No errors to be displayed in the logs.

**Logs**
Shows that the release works, then shows errors during what I assume is the dry-run then after bumping image version it working again.

```
ts=2019-12-03T19:58:19.28044721Z caller=operator.go:309 component=operator info="enqueuing release" resource=system:helmrelease/sealed-secrets-controller
I1203 19:58:19.364075       6 upgrade.go:79] preparing upgrade for sealed-secrets-controller
I1203 19:58:19.535110       6 upgrade.go:87] performing update for sealed-secrets-controller
I1203 19:58:19.544234       6 upgrade.go:225] creating upgraded release for sealed-secrets-controller
I1203 19:58:19.563804       6 client.go:149] checking 9 resources for changes
I1203 19:58:19.567442       6 client.go:384] Looks like there are no changes for ServiceAccount "sealed-secrets-controller"
I1203 19:58:19.575723       6 client.go:384] Looks like there are no changes for ClusterRole "secrets-unsealer"
I1203 19:58:19.590303       6 client.go:384] Looks like there are no changes for Role "sealed-secrets-controller-key-admin"
[...]

PR #131, opened 2019-12-03
Setting ClientOnly when performing the dry-run during shouldSync

stefansedich:
After discussions in Slack with @hiddeco about #130 it was suggested that when doing the dry-run instlal to see if we should sync to set ClientOnly so that we skip any server side validation.

This appears to work fixing both my issues with sealed-secrets and validation issues I had with cert-manager. I no longer see any errors and was able to successfully do a fresh release install and upgrade.

I was unable to find any such option for helm2 so left this out at this stage, and if anyone wants to test this out I have pushed it to `docker.io/stefansedich/helm-operator:bug-130-16b75226-wip`.

Note: There are some fmt changes that were applied, I run gofmt on save so if I have something setup wrong let me know.

### DIFF
--- a/pkg/helm/upgrade.go
+++ b/pkg/helm/upgrade.go
@@ -15,6 +15,7 @@ type UpgradeOptions struct {
 	Install      bool          `json:"install,omitempty"`
 	DisableHooks bool          `json:"disableHooks,omitempty"`
 	DryRun       bool          `json:"dryRun,omitempty"`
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM

stefansedich:
Thanks @stefanprodan! I just rolled out `helm-v3-dev-53b6a21d` and so far so good, will monitor how things go over time.

PR #139, opened 2019-12-10
Add Service and ServiceMonitor templates to the chart

sa-spag:
Hello 🤝 

This adds a [ServiceMonitor](https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/running-exporters.md) template to integrate with the [Prometheus Operator](https://github.com/coreos/prometheus-operator). Pretty much the same as fluxcd/flux#2668, plus a Service template which was missing.


stefanprodan (approved):
LGTM

PR #124, opened 2019-11-26
helm/v3: import repositories from file

hiddeco:
This commits adds the base for importing repositories from a file for
Helm v3, and other tools around Helm v3 repository management. The
implemented functionalities are still limited, but sufficient to
resolve the reported `no repository definition` error bug.

Fixes #68.

PR #141, opened 2019-12-10
Support import of repositories for Helm v2 _and_ v3

hiddeco:
This PR adds the missing bits in the `helm/v2` package, removes the hardcoded import that was temporary put into place in #124, and replaces it with a dedicated `--helm-repository-import` flag which accepts a string slice of `[helm version]:[path]` pairs.

### DIFF
--- a/pkg/helm/v2/repository.go
+++ b/pkg/helm/v2/repository.go
@@ -1,0 +1,142 @@
+package v2
+
+import (
+	"os"
+	"sync"
+
+	"github.com/pkg/errors"
+
+	"k8s.io/helm/pkg/getter"
+	"k8s.io/helm/pkg/repo"
+)
+
+var (
+	repositoryConfigLock sync.RWMutex
+	getters              = getter.Providers{{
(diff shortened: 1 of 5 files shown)


stefanprodan (approved):
LGTM

pkg/helm/v2/repository.go
```diff
@@ -0,0 +1,142 @@
...
+	"sync"
+
+	"github.com/pkg/errors"
+
+	"k8s.io/helm/pkg/getter"
+	"k8s.io/helm/pkg/repo"
+)
+
+var (
+	repositoryConfigLock sync.RWMutex
```
2opremio, line 14: is there a good reason to have this as a global variable? Can't it be part of `Helm2`?

hiddeco: There can theoretically be multiple versions of `Helm2` while they all speak with the same repository index file (this is not completely true, as they can have different `HELM_HOME`s and are thus working with a different file, but this is the simplest secure option).

PR #148, opened 2019-12-13
e2e: test git and chart repository rollbacks

hiddeco:
This has brought to my attention that we currently do not take validation errors into account for Helm v3, we do not want to perform a rollback on those as they happen pre-upgrade.

---

The above has been taken care of, and tests have been added to address part of #103.


stefanprodan (approved):
LGTM

PR #150, opened 2019-12-16
e2e: parallelize end-to-end tests

hiddeco:


2opremio:
You are missing setting the number of clusters when invoking `make e2e` from CircleCI

hiddeco:
@2opremio I think you were looking at an old commit.

.circleci/config.yml
```diff
@@ -52,7 +53,16 @@ jobs:
...
+      - run:
+          name: End-to-end Helm v2
+          command: E2E_KIND_CLUSTER_NUM=4 make e2e
+          environment:
+            HELM_VERSION: v2
+      - run:
+          name: End-to-end Helm v3
+          command: E2E_KIND_CLUSTER_NUM=4 make e2e
+          environment:
+            HELM_VERSION: v3
```
2opremio, line 65: Does it make sense to both pass the env variables through the shell and through `environment`? I would go for one of the two mechanisms

2opremio:
> @2opremio I think you were looking at an old commit.

I think you may be right, it looks good now

.circleci/config.yml
hiddeco: I used the `environment` for the Helm version to highlight the difference between the two `run` items.

2opremio: alright

2opremio (approved):

PR #171, opened 2020-01-03
helm/v3: include credentials when pulling chart

PaulFarver:
Using a private helm repository currently fails, when using helm v3. The issue arises, when attempting to download the chart itself. The credentials are not passed to the downloader.

This pull request rectifies this issue, by passing the credentials on, in the same manner as this [similar code in the helm project](https://github.com/helm/helm/blob/8dddb18de723449282fde2c4266e4ae9263a50e0/pkg/action/pull.go#L59) in the PullWithRepoURL func

### DIFF
--- a/pkg/helm/v3/pull.go
+++ b/pkg/helm/v3/pull.go
@@ -2,6 +2,7 @@ package v3
 
 import (
 	"helm.sh/helm/v3/pkg/downloader"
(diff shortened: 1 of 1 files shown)


hiddeco (approved):
Thanks a lot @PaulFarver! I rebased the branch on `master` and added a small addition to also include certificate references. Good to go! :1st_place_medal: :star:

Issue #94, opened 2019-10-31
Git over HTTPS support


masterkain:
Helm Operator allows for recycling the git ssh key eventually used in a Flux setup, however Flux does also support Git over HTTPS to fetch its state:
https://github.com/fluxcd/flux/blob/master/chart/flux/README.md#flux-with-git-over-https
which is the method I initially adopted.

However as per Helm Operator's readme the only method supported is the git deploy key:
https://github.com/fluxcd/helm-operator/blob/master/chart/helm-operator/README.md#use-fluxs-git-deploy-key

Is using `flux-git-auth` instead of `flux-git-deploy` supported somehow? In any case how would the `spec.chart.git` entry look like in our HelmRelease?

jbuettnerbild:
i have the same problem. i don't want to configure flux and the helm operator differently. One time with git username and token and otherwise with a ssh key.

mmaask:
This should be supported indeed. Currently helm operator git clones are all getting access denied.

hiddeco:
The only way this would work at the moment is if one would provide plain credentials in the `HelmRelease` `spec.chart.git` entry. The reason this is not required for Flux is because Flux is able to utilize the [variable substitution built in to Kubernetes](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config), something that is not possible using custom resources.

To make this possible without leaking credentials in a `HelmRelease` resource we need to come up with a secure and straight forward way to provide a reference to a secret containing the credentials, and add them to the parsed `spec.chart.git` URL before we request a new Git mirror.

This will likely also include some brain gymnastics around mirror names as this equals to the `spec.chart.git` value at the moment to re-use git mirrors cross releases. To prevent leaking access across `HelmRelease` resources and/or namespaces this would need a reference to the credentials being referred to and/or namespace the resource is in (or e.g. a simple checksum of all values combined).

chancez:
Could you not just support configuring helm-operator with a single set of credentials and simply use them during cloning if present? I think supporting creds per-HelmRelease CR is a nice-to-have, but being able to simply specify HTTPS credentials (via env) to helm-operator would suffice for many use-cases. In fact, you could probably even just use `os.ExpandEnv` on the URI and call it a day, and we could then use env vars in the URI and it would expand them the way anyone expects.

hiddeco:
[...]

Issue #142, opened 2019-12-10
Manage Helm chart repositories using custom resources


hiddeco:
**Edit**: This has been implemented in helm-controller the successor of helm-operator, please see https://toolkit.fluxcd.io/

In #124 and #141 boilerplate was added to manage Helm chart repositories for Helm `v2` and `v3`. The current implementation is however extremely limited and still requires the user to provide (and manage) a repository index file.

It would be more user friendly, and declarative, to introduce a new custom resource definition called i.e. `HelmChartRepository` to make it possible to manage the repositories using Kubernetes resources.

stefanprodan:
**Proposal: decoupling chart/git repositories from HelmReleases**

Features:
* a release references a chart source (can be helm chart repo or git repo) 
* a chart source references a secret (helm chart repo basic auth, git https basic auth or git ssh key)
* a release can reference a chart source defined in a different namespace so that cluster admins can manage the sources in a restricted namespace
* by running multiple helm-op instances we can restrict with RBAC what sources an operator instance can use
* by allowing cross-namespace references, we can avoid duplicating the source definitions (e.g. stable) for each namespace 

Chart repository with basic auth:

```yaml
apiVersion: helm.fluxcd.io/v1
kind: ChartRepository
metadata:
  name: azurecr
  namespace: fluxcd
spec:
  url: https://<repository>.azurecr.io/helm/v1/repo #required
  secretRef: #optional
    name: azurecr-auth #required
```

Chart repository basic auth secret format:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: azurecr-auth
  namespace: fluxcd
type: Opaque
data:
  username: <BASE64> #required
  password: <BASE64> #required
  caFile: <BASE64> #optional (path to file inside the helm-op container)
  certFile: <BASE64> #optional (path to file inside the helm-op container)
  keyFile: <BASE64> #optional (path to file inside the helm-op container)
```

Reference ChartRepository in HelmRelease with cross-namespace support:

```yaml
apiVersion: helm.fluxcd.io/v2
kind: HelmRelease
metadata:
  name: podinfo
  namespace: demo
spec:
  chart:
    name: podinfo #required (either name or path)
    version: 3.1.0 #required (when using chartRepositoryRef)
    chartRepositoryRef: #required (either chartRepositoryRef or gitRepositoryRef) 
      name: azurecr #required
      namespace: fluxcd #optional (defaults to HelmRelease ns)
```

[...]

PR #172, opened 2020-01-03
chartsync: support Git HTTPS credentials using secret

richardcase:
Started to add https basic authentication to the git chart source. the
username and password can be set as a static value or from a secret.
For example:

```yaml
spec:
  releaseName: my-app
  chart:
    git: https://github.com/someonesomerepo.git
    ref: master
    path: charts/my-app
    secretRef:
      name: git-credentials

```

Still a work in progress

Relates to: #94


stefanprodan:
@richardcase I think this should be paused until we decide how to handle git/helm repos outside the HelmRelease spec. See https://github.com/fluxcd/helm-operator/issues/142

richardcase:
> @richardcase I think this should be paused until we decide how to handle git/helm repos outside the HelmRelease spec. See #142

@stefanprodan  - Sure. Although i really need this functionality for a specific scenario i'm working on. I'll work from a fork. If you need any help with #142 let me know

stefanprodan:
If you need this then please make it more compact like in my proposal:

```yaml
spec:
  releaseName: my-app
  chart:
    git: https://github.com/someonesomerepo.git
    ref: master
    path: charts/my-app
    secretRef: #optional
      name: git-basic-auth #required
```

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: git-basic-auth
type: Opaque
data:
  username: <BASE64> #required
  password: <BASE64> #required
```

richardcase:
> If you need this then please make it more compact like in my proposal

Sure i can do that. I will make that change now

hiddeco:
In addition to all of the above, instead of merging in `master` it is better for reviewers and Git history clutter to rebase on top of `master`.

richardcase:
> In addition to all of the above, instead of merging in `master` it is better for reviewers and Git history clutter to rebase on top of `master`.

Understood, that was a mess up my side....a case of PEBKAC

richardcase:
@stefanprodan @hiddeco  - i've made the change to simplify the specifiying of the creds as suggested. So as suggested you have:

```yaml
spec:
  releaseName: my-release
  chart:
    git: https://github.com/someorg/somerepo.git
    ref: master
    path: charts/my-chart
    secretRef:
      name: git-pwd
```
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: git-pwd
type: Opaque
data:
  password: supersecret
  username: joeblogs
```

cmd/helm-operator/main.go
```diff
@@ -23,8 +23,8 @@ import (
 	clientset "github.com/fluxcd/helm-operator/pkg/client/clientset/versioned"
 	ifinformers "github.com/fluxcd/helm-operator/pkg/client/informers/externalversions"
 	"github.com/fluxcd/helm-operator/pkg/helm"
-	"github.com/fluxcd/helm-operator/pkg/helm/v2"
-	"github.com/fluxcd/helm-operator/pkg/helm/v3"
+	v2 "github.com/fluxcd/helm-operator/pkg/helm/v2"
+	v3 "github.com/fluxcd/helm-operator/pkg/helm/v3"
```
hiddeco: The alias here is obsolete, and is probably caused by some enabled linter.

pkg/apis/helm.fluxcd.io/v1/types.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/ghodss/yaml"
 
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
hiddeco: These are duplicates (`corev1` should be kept).

pkg/chartsync/git.go
```diff
@@ -3,16 +3,22 @@ package chartsync
...
 	"github.com/go-kit/kit/log"
 
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
 
-	"github.com/fluxcd/helm-operator/pkg/apis/helm.fluxcd.io/v1"
+	v1 "github.com/fluxcd/helm-operator/pkg/apis/helm.fluxcd.io/v1"
```
hiddeco: Same as the above.

pkg/chartsync/git.go
```diff
@@ -3,16 +3,22 @@ package chartsync
...
+	"net/url"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/fluxcd/flux/pkg/git"
 	"github.com/go-kit/kit/log"
 
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
hiddeco, line 16: :+1:

pkg/chartsync/git.go
```diff
@@ -281,12 +291,22 @@ func (c *GitChartSync) sync(hr *v1.HelmRelease, mirrorName string, repo *git.Rep
 // maybeMirror requests a new mirror for the given remote. The return value
 // indicates whether the repo was already present (`true` if so,
 // `false` otherwise).
-func (c *GitChartSync) maybeMirror(mirrorName string, remote string) bool {
+func (c *GitChartSync) maybeMirror(mirrorName string, source *v1.GitChartSource, namespace string) bool {
+	gitURL := source.GitURL
+	if source.SecretRef != nil {
```
hiddeco: As you are doing the same `nil` check in `addAuthForHTTPS`, I would reduce this to

```go
gitURL := source.GitURL
var err error
if gitURL, err = c.addAuthForHTTPS(source.GitURL, source.SecretRef, namespace); err != nil {
  c.logger.Log("error", GitAuthError{err}.Error())
  return false
}
```

hiddeco (commented):
Besides the in-code comments, I think the `mirrorName` function needs to be made smarter as the credentials from other `HelmRelease` resources (and namespaces) will now leak by configuring a `source.GitURL` (without credentials) that is already in use by another `HelmRelease` (with credentials).

You may get away with just adding the namespace, if I am not mistaken this is the same approach Flux uses for image pull secrets.

cmd/helm-operator/main.go
richardcase: Changed this.

pkg/chartsync/git.go
richardcase: @hiddeco - every time i removed the v1 alias it adds it back so this hasn't been resolved i'm afraid

pkg/apis/helm.fluxcd.io/v1/types.go
richardcase: Changed

pkg/chartsync/git.go
richardcase: A good suggestion, changed.

richardcase:
> You may get away with just adding the namespace, if I am not mistaken this is the same approach Flux uses for image pull secrets.

@hiddeco - i checked and flux does use the namespace. I changed the `mirrorName` to use a composite of namespace, auth secret name, git url. Do you think this is ok?

pkg/chartsync/git.go
hiddeco: I will remove it for you before the merge, no worries.

pkg/chartsync/git.go
```diff
@@ -314,7 +332,60 @@ func (c *GitChartSync) helmReleasesForMirror(mirror string) ([]*v1.HelmRelease,
 // per namespace, per auth.
 func mirrorName(hr *v1.HelmRelease) string {
 	if hr != nil && hr.Spec.GitChartSource != nil {
-		return hr.Spec.GitURL
+		secretName := "noauth"
```
hiddeco: I think this approach is ideologically correct when secrets are being used, but this also means we will do a lot more (duplicate) polling for e.g. releases without HTTPS authentication. My propose would be to change it to something like this:

```go
if hr != nil && hr.Spec.GitChartSource != nil {
	if hr.Spec.GitChartSource.SecretRef == nil {
		return hr.Spec.GitURL
	}
	return fmt.Sprintf("%s/%s/%s", hr.GetNamespace(), hr.Spec.GitChartSource.SecretRef.Name, hr.Spec.GitURL)
}
return ""
```

In addition, the `TODO(michael)` comment can also be removed.

hiddeco:
@richardcase thanks a lot :1st_place_medal: 

@stefanprodan can you take a look and see if this matches your expectations for the issue you mentioned? Feature wise it looks fine and mergeable to me (after I get rid of the persistent alias caused by the linter).

stefanprodan:
@hiddeco it looks Ok to me, it needs a rebase with master are you going to do that when fixing the aliases?

hiddeco:
@stefanprodan yep, will do the rebase and some minor fixes that triggered my OCD :+1:

hiddeco (approved):
Rebased and tidied up, awesome contribution @richardcase :raised_hands: :cake:

pkg/chartsync/git.go
```diff
@@ -284,12 +292,20 @@ func (c *GitChartSync) sync(hr *v1.HelmRelease, mirrorName string, repo *git.Rep
...
+		return false
+	}
+
 	ok := c.mirrors.Mirror(
-		mirrorName, git.Remote{URL: remote}, git.Timeout(c.config.GitTimeout),
+		mirrorName, git.Remote{URL: gitURL}, git.Timeout(c.config.GitTimeout),
 		git.PollInterval(c.config.GitPollInterval), git.ReadOnly)
 	if !ok {
-		c.logger.Log("info", "started mirroring new remote", "remote", remote, "mirror", mirrorName)
+		c.logger.Log("info", "started mirroring new remote", "remote", source.GitURL, "mirror", mirrorName)
```
stefanprodan, line 308: Do we log the credentials here?

pkg/chartsync/git.go
```diff
@@ -78,11 +85,12 @@ func (c sourceRef) forHelmRelease(hr *v1.HelmRelease) bool {
 }
 
 func NewGitChartSync(logger log.Logger,
-	lister lister.HelmReleaseLister, cfg GitConfig, queue ReleaseQueue) *GitChartSync {
+	coreV1Client corev1client.CoreV1Interface, lister lister.HelmReleaseLister, cfg GitConfig, queue ReleaseQueue) *GitChartSync {
 
 	return &GitChartSync{
 		logger:             logger,
 		config:             cfg,
+		coreV1Client: 		coreV1Client,
```
stefanprodan: Please run `go fmt`.

pkg/chartsync/git.go
hiddeco: No, `source.GitURL` does not include the credentials but is the 1:1 copy from the `HelmRelease` definition.

stefanprodan: Great!

stefanprodan:
In the future we should consider adding support for cross-namespace secrets like we have done in https://github.com/fluxcd/helm-operator/pull/219

stefanprodan (approved):
LGTM

Issue #164, opened 2019-12-27
Chart lookup fails and appears to be looking in the wrong cache


palemtnrider:
**Describe the bug**
Using 0.4.0 (1.0.0-rc5) and the podinfo hr doesn't load with the following error:
`chart unavailable: no cached repo found. (try 'helm repo update'). open /var/fluxd/helm/repository/cache/stable-index.yaml: no such file or directory`

``` shell
cat <<EOF | kubectl apply -f -
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  name: podinfo
  namespace: default
spec:
  releaseName: podinfo
  chart:
    repository: https://stefanprodan.github.io/podinfo
    version: 2.1.0
    name: podinfo
  values:
    replicaCount: 1
EOF
```

Switching back to 0.3.0 and the chart is loaded correctly.  

**To Reproduce**
Steps to reproduce the behaviour:
0. What's your setup?
helm 2
helm-operator chart 0.4.0
minikube
`helm install --wait --name helm-operator \
--namespace fluxcd \
fluxcd/helm-operator `
``` shell
helm upgrade -i helm-operator fluxcd/helm-operator \
--namespace fluxcd \
--set configureRepositories.enable=true \
--set configureRepositories.repositories[0].name=stable \
--set configureRepositories.repositories[0].url=https://kubernetes-charts.storage.googleapis.com \
--set configureRepositories.repositories[1].name=podinfo \
--set configureRepositories.repositories[1].url=https://stefanprodan.github.io/podinfo
```
2. Click on 
add hr
``` shell
cat <<EOF | kubectl apply -f -
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  name: podinfo
  namespace: default
spec:
  releaseName: podinfo
  chart:
    repository: https://stefanprodan.github.io/podinfo
    version: 2.1.0
    name: podinfo
  values:
    replicaCount: 1
EOF
```
3. Scroll down to 
`kubectl describe hr podinfo`

**Expected behavior**
The podinfo chart to be deployed or the error to reference the podinfo cache not the stable cache

**Logs**
If applicable, please provide logs. In a standard stand-alone installation, you'd get this by running `kubectl logs -n default deploy/flux-helm-operator`.

**Additional context**
Switching back to 0.3.0 fixed the issue.

See slack discussion at https://weave-community.slack.com/archives/C4U5ATZ9S/p1577392027025900

[...]

PR #173, opened 2020-01-06
chart: configure Helm v3 repositories

PaulFarver:
Currently when deploying the helm chart with v3 enabled and a repositories file is configured with configureRepositories, the repositories are only available to helm2, since helm3 has a different config path

This pull request adds logic to the deployment template to only configure repositories for the enabled versions of helm.

Addresses part of #186

### DIFF
--- a/chart/helm-operator/templates/deployment.yaml
+++ b/chart/helm-operator/templates/deployment.yaml
@@ -136,11 +136,19 @@ spec:
         {{- end }}
         {{- if .Values.configureRepositories.enable }}
         - name: {{ .Values.configureRepositories.volumeName | quote }}
-          mountPath: /var/fluxd/helm/repository/repositories.yaml
+          mountPath: /root/.helm/repository/repositories.yaml
           subPath: repositories.yaml
           readOnly: true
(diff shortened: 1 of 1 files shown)


stefanprodan:
The operator should copy the repositories file for each version on its own, I guess this is bug? @hiddeco can you confirm? relates to #164

PaulFarver:
My initial approach was actually to include the `--helm-repository-import` for every enabled version, but since the file is already mounted to the helm v2 config directory it seemed to be more in line with the design to just mount it to the helm v3 config directory as well.

hiddeco:
The changes in this PR are technically fine, given the new setup however, I think it would be wiser to mount the file outside of the expected (Helm) location and supply it to the `--helm-repository-import` flag, so that the operator takes the proper route and steps to import all repositories and update the indexes (the latter being the most important).

PaulFarver:
@hiddeco I have switched to a `--helm-repository-import` approach, but the logic taking care of the repository cache is still there in the same way as before.

hiddeco (approved):
Tested and confirmed working, thanks a lot @PaulFarver :stars:

PR #177, opened 2020-01-07
Update Helm packages to 3.0.2 and 2.16.1

stefanprodan:
In Helm 3.0.1 the chart deps were broken, updating to 3.0.2 to pull in  https://github.com/helm/helm/pull/7009

Note that Helm 3.0.2 has a regression bug and CRDs can't be upgraded, a fix will be shipped with 3.0.3 https://github.com/helm/helm/pull/7269


go.sum
```diff
@@ -696,6 +700,8 @@ gotest.tools v2.2.0+incompatible h1:VsBPFP1AI068pPrMxtb/S8Zkgf9xEmTLJjfM+P5UIEo=
 gotest.tools v2.2.0+incompatible/go.mod h1:DsYFclhRJ6vuDpmuTbkuFWG+y2sxOXAzmJt81HFBacw=
 helm.sh/helm/v3 v3.0.1 h1:gEs30kweCOnLFK9Diq2S8b+VHmWQ2oi465GhqTc3ZxI=
 helm.sh/helm/v3 v3.0.1/go.mod h1:sI7B9yfvMgxtTPMWdk1jSKJ2aa59UyP9qhPydqW6mgo=
```
hiddeco: Can you explain this version being kept?

stefanprodan: Fixed!

hiddeco (approved):
:100: thanks Stefan

hiddeco (approved):
~On second thoughts, I think `docker/helm2.version` should also be updated to stay in sync with the Go dep.~

Already the case, Go dep was behind image dep.

PR #178, opened 2020-01-07
e2e: fix flakiness of various tests

hiddeco:
This improves the stability of the end-to-end tests by fixing the flakiness of various tests. In addition, to help (new) contributors with the setup of the end-to-end tests a check has been added to confirm the `parallel` version on the system is [GNU Parallel and not Tollef's `parallel`, as they behave differently](https://www.gnu.org/software/parallel/parallel_design.html#tollef).

### DIFF
--- a/test/e2e/run.bash
+++ b/test/e2e/run.bash
@@ -44,13 +44,13 @@ if ! kubectl version > /dev/null 2>&1; then
 
   echo '>>> Creating Kind Kubernetes cluster(s)'
   KIND_CONFIG_PREFIX="${HOME}/.kube/kind-config-${KIND_CLUSTER_PREFIX}"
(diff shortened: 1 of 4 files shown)


stefanprodan:
@hiddeco fix this typo also `Git mituation causes upgrade`

stefanprodan (approved):

test/e2e/run.bash
```diff
@@ -37,15 +37,20 @@ E2E_KIND_CLUSTER_NUM=${E2E_KIND_CLUSTER_NUM:-1}
 if ! kubectl version > /dev/null 2>&1; then
   install_kind
 
+  # We require GNU Parallel, but some systems come with Tollef's parallel (moreutils)
+  if ! parallel -h | grep -q "GNU Parallel"; then
```
2opremio, line 41: We should include this in Flux too

Issue #179, opened 2020-01-07
e2e test failure on git clone


richardcase:
**Describe the bug**
When running the e2e tests i am getting the following error:

```bash
# Cloning into '/var/folders/wx/6t6918v113xdxvvy1t9ybq1h0000gn/T/tmp.ezdDlz3J'...
# Warning: Permanently added '[localhost]:51964' (ECDSA) to the list of known hosts.
# Received disconnect from 127.0.0.1 port 51964:2: Too many authentication failures
# Disconnected from 127.0.0.1 port 51964
# fatal: Could not read from remote repository.
#
# Please make sure you have the correct access rights
# and the repository exists.
#
```

This is when running the Git mituation causes upgrade test in `15_upgrade.bats`

**To Reproduce**
Run the following:
```bash
make e2e
```

**Expected behavior**
The tests to pass (i.e. git clone to succeed in this case)

PR #180, opened 2020-01-07
e2e: fix authentication issue on `git clone`

richardcase:
In the `15_upgrade` tests when it was doing a git clone it would
report `Too many authentication failures` if there are a number
of different ssh keys on the system. As the tests use a generated
key pair we can tell ssh to only use the supplied identity file and
not try all found keys.

Relates to: #179


hiddeco (approved):
Thank you @richardcase. Much appreciated that you took the time to get to the bottom of this :tulip:

PR #182, opened 2020-01-07
release: detect change of charts base values

### LINKED ISSUE #174: Changing values.yaml in chart does not result in new deployment
**Describe the bug**
When changing a value inside the values.yaml for a referenced chart it appears as this change is not picked up and results in a new deployment not happening.

Given the following release:

```
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  name: podinfo-dev
  namespace: default
  annotations:
    fluxcd.io/automated: "true"
    filter.fluxcd.io/chart-image: glob:dev-*
spec:
  releaseName: podinfo-dev
  chart:
    git: https://github.com/stefansedich/helm-operator-get-started
    path: charts/podinfo
    ref: master
  values: {}
```

If I were to then go and update the `replicaCount` in the charts values.yaml I would expect a deployment to happen but after a number of sync cycles nothing happens. If I were to go change something inside say `templates/deployment.yaml` however this does get picked up.

**To Reproduce**
1. Deploy the above release (you might want to deploy pointing to a fork you can edit)
2. Change replicaCount to another value
3. Wait 5 minutes
4. Witness no deployment has happened updating replica count

**Expected behavior**
I would expect a change in values.yaml would result in a release to happen, especially in this case where it is something like number of replicas changing.

**Logs**
Nothing special shown in the logs, just normal sync happening.

**Additional context**
- Helm operator version: rc5
- Targeted Helm version: 3.0.1
- Kubernetes version: 1.14.6
- Git provider: GitHub
- Container registry provider: N/A

stefansedich:
Fix #174 

From my investigation I found that upon running the dry-run we compare to see if the release values have changed but never actually checked to see if the charts base values themselves actually changed.

This adds a checksum of the chart values to the Chart type which in turn leads to the chart comparison to see a difference resulting in a release to occur when the base charts values are changed.

I also did some slight refactoring to move the existing Checksum logic into the helm package and changed any existing stuff to use that.

@hiddeco let me know what other changes you envisage here, one thing we could do and I am unsure about is return the actual values on the Chart and not just a checksum, but given we are not using them for anything else the checksum might be the simplest route.

### DIFF
--- a/pkg/apis/helm.fluxcd.io/v1/types.go
+++ b/pkg/apis/helm.fluxcd.io/v1/types.go
@@ -8,12 +8,11 @@ import (
 	"github.com/fluxcd/flux/pkg/resource"
 	"github.com/ghodss/yaml"
 
-	"k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"k8s.io/helm/pkg/chartutil"
-
-	"github.com/fluxcd/helm-operator/pkg/helm/v2"
+	"github.com/fluxcd/helm-operator/pkg/helm"
+	helmv2 "github.com/fluxcd/helm-operator/pkg/helm/v2"
 )
 
 // +genclient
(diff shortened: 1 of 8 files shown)


hiddeco (commented):
Awesome! I wrote an end-to-end test to confirm it works.

Small comment about import aliases. And  to keep the history clean, can you please smash your commits into one descriptive commit and rebase the branch instead of merging in master? Thanks :star:

stefansedich:
@hiddeco should be rebased now, my bad I am so used to using the squash feature in GitHub now I often forget to rebase PRs.

As far as the comment on import aliases I did not see any comment? I am not doing anything special myself there I assume it is vim-go doing it for me as it was something it updated upon saving/fmt the file for me.

Thanks for the test too I was going to talk to you about this, but this gives me a good guide for the future on e2e tests.

hiddeco (approved):

PR #183, opened 2020-01-08
chart: support configuration of cert files for repositories

### LINKED ISSUE #168: Chart issue - configureRepositories.repositories[0].caFile is not settable
**Describe the bug**

It is not possible to set `caFile`, `certFile` or `keyFile` because the template function will always set them to `""`.

**To Reproduce**

Run `helm install --set configureRepositories.repositories[0].enable=true --set configureRepositories.repositories[0].caFile="something"` and it will never be set

**Expected behavior**

`caFile` should be a settable value. Not be forced to `""`

**Additional context**

[Permalink to relevant code](https://github.com/fluxcd/helm-operator/blob/843e8afd7ae6dfdf08be8161dddb5c3953fb6124/chart/helm-operator/templates/_helpers.tpl#L63-L73
)

- Helm operator version:  Helm operator 1.0.0-rc4, chart 0.4.0
- Kubernetes version: 1.16.3

hiddeco:
Fixes #168

### DIFF
--- a/chart/helm-operator/templates/_helpers.tpl
+++ b/chart/helm-operator/templates/_helpers.tpl
@@ -64,9 +64,9 @@ repositories:
 - name: {{ required "Please specify a name for the Helm repo" .name }}
   url: {{ required "Please specify a URL for the Helm repo" .url }}
   cache: /var/fluxd/helm/repository/cache/{{ .name }}-index.yaml
-  caFile: ""
-  certFile: ""
-  keyFile: ""
+  caFile: "{{ .caFile | default "" }}"
+  certFile: "{{ .certFile | default "" }}"
+  keyFile: "{{ .keyFile | default "" }}"
   password: "{{ .password | default "" }}"
   username: "{{ .username | default "" }}"
 {{- end }}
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

weaktyper:
@hiddeco Thanks !

weaktyper:
@hiddeco Having experimented with this change, I have found it is not working. I think the issue is probably the outer `"`

hiddeco:
@weaktyper

What was the `values.yaml` you supplied to Helm (or command line arguments), and did the operator (or Helm) give you any errors back?

weaktyper:
supplied command line arguments and also mounted the `cacert` to a filepath using an extravolume/extravolumemount and verified it in the container - the container filepath is 100% correct and so too is the certificate content. Verified the same `cacert` works outside the k8s cluster with just plain helm on laptop.

error message: 
```
component=helm version=v2 error="unable to get an update from the chart repository" url=https://<your_domain>/chartrepo/private err="Get https://<you_domain>/chartrepo/private/index.yaml:
x509: certificate signed by unknown authority (possibly because of \"crypto/rsa: verification error\" while trying to verify candidate authority certificate \"<your_ca_name>\")"
```

Issue #8, opened 2019-05-22
Helm 3 support


cdenneen:
With the release of Helm 3 would like to track progress for this integration

stealthybox:
> There is also a question about how the operator is going to know what helm version it needs to call, introducing a version field to the HelmRelease spec is an option (which would default to v2 for backwards compatibility), and this would give you the option to migrate your releases one-by-one (by changing the version field to v3 one release at a time).

The `Chart.yaml` has an APIVersion that indicates use of Helm v3+.
We can use that to determine what version of helm the Chart requires.
https://v3.helm.sh/docs/faq/#chart-yaml-apiversion-bump

This requires less user-action and removes edge cases where the wrong client version is set for a Chart that is not supported.
For transparency we should consider logging and adding the helmClient version used to the Status object.

> I see room here for dropping the .spec.releaseName field and always use the name of a HelmRelease as a release name.

+1, this should be the v3 default, even if we still support `HR.Spec.ReleaseName`.

> The rationale for .spec.releaseName is to account for "taking over" a release -- that is, introducing a HelmRelease resource to manage an existing release, and thereby avoid disruption. That use case hasn't gone away, though we could in principle decide not to care about it.

Release names used to be cluster-global, but now they are now fully namespaced in helmv3.
This makes this use-case much less useful, because nothing stops you from taking over a release by just adding a HelmRelease of the same name in the same Namespace.

If we still support `HR.Spec.TargetNamespace` for helmv3, then I can see `HR.Spec.ReleaseName` being useful.
This would be a value-add of helm-operator since you could manage multiple releases of the same name in different Namespaces from HR's in the same management Namespace.
These would otherwise have a name collision unless you overrode `HR.Spec.ReleaseName` on at least `n-1` of them.

stealthybox:
Thinking about U/X in the end:
We may want a global fluxd config option to enable/disable use helmv2/v3.
This is so users can enforce policies for not using v2 Tiller or beta v3 features.

gsf:
I was wondering whether there was progress going on outside of this issue. Looks like #42 and #55 have both been merged into the ongoing helm-v3-dev branch.

hiddeco:
@gsf `helm-v3-dev` is the branch you want to follow for updates on the Helm 3 support.

[...]

PR #185, opened 2020-01-08
release: filter out nil release resources during parsing

hiddeco:
This was once fixed in #47, but returned as a regression bug due
to a merge mistake.

Fix for https://github.com/fluxcd/helm-operator/issues/8#issuecomment-569433618


stefanprodan (approved):
LGTM

PR #187, opened 2020-01-08
Release v1.0.0-rc6

hiddeco:


stefanprodan (approved):
LGTM 🚀

davidholsgrove:
Can we bump the Chart version as well as appversion to avoid differently cached copies of the chart values?

stefanprodan:
@davidholsgrove  the chart publishing comes later, here is our release process https://github.com/fluxcd/helm-operator/blob/master/internal_docs/releasing.md

PR #190, opened 2020-01-08
release: ignore manifest in annotator if we fail to unmarshal

stefansedich:
#185 covered the case where the len(bytes)==0 however in my testing I witnessed that the reason I still get this error is because I am seeing manifests come through that are simply the string "null".

### DIFF
--- a/pkg/release/annotator.go
+++ b/pkg/release/annotator.go
@@ -14,7 +14,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
-	"github.com/fluxcd/helm-operator/pkg/apis/helm.fluxcd.io/v1"
+	v1 "github.com/fluxcd/helm-operator/pkg/apis/helm.fluxcd.io/v1"
 	"github.com/fluxcd/helm-operator/pkg/helm"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/release/annotator.go
```diff
@@ -14,7 +14,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
-	"github.com/fluxcd/helm-operator/pkg/apis/helm.fluxcd.io/v1"
+	helmfluxcdiov1 "github.com/fluxcd/helm-operator/pkg/apis/helm.fluxcd.io/v1"
```
stefansedich: Changed to be consistent with other modules as my linter was adding the v1 again.

stefansedich:
After talking with @hiddeco going to dive in and try understand where the "null" is actually coming from.

pkg/release/annotator.go
```diff
@@ -89,19 +89,9 @@ func releaseManifestToUnstructured(manifest string, logger log.Logger) []unstruc
...
-
-		if len(bytes) == 0 {
-			continue
-		}
-
 		var u unstructured.Unstructured
-		if err := u.UnmarshalJSON(bytes); err != nil {
-			logger.Log("err", err)
+
+		if err := yaml.Unmarshal([]byte(manifest), &u); err != nil {
```
stefansedich, line 94: Change to just attempt unmarshalling directly to the Unstructured type and ignore if there are any issues.

hiddeco (approved):
Thanks for getting to the bottom of this @stefansedich :cake:

PR #192, opened 2020-01-09
pkg: update Flux package to v1.17.0 kube-1.16 branch

stefanprodan:
- pin Flux to kube-1.16 branch to break weaveworks/flux circular dependency (to be removed on Flux 1.18)


go.mod
```diff
@@ -3,23 +3,20 @@ module github.com/fluxcd/helm-operator
 go 1.13
 
 require (
-	github.com/bugsnag/bugsnag-go v1.5.3 // indirect
+	github.com/Masterminds/semver v1.5.0 // indirect
 	github.com/fluxcd/flux v1.15.0
```
hiddeco: Shouldn't this also be updated?

stefanprodan: It's an indirect, I can pin it but I see no point in doing so.

hiddeco (approved):

PR #196, opened 2020-01-10
helm/v3: prevent spurious upgrades due to missing dependencies

### LINKED ISSUE #194: Charts with dependencies are upgraded every cycle
**Describe the bug**
When deploying a helm chart with other charts as dependencies, the helm operator upgrades on every sync cycle.
if `--log-release-diffs` is enabled, the helm operator reports that the `chart has diverged`, and logs the diff. This diff is always, that the dependencies have been added.

I tried to debug a little myself, and it seems that the release object returned from: https://github.com/fluxcd/helm-operator/blob/ea87d4765aa67eb3cfd5751cefbfcf33ae0d197e/pkg/helm/v3/status.go#L22 does not have any dependencies attached. Running `.Dependencies()` on it returns an empty slice

**To Reproduce**
1. Apply a helmrelease of a chart with one or more dependencies (example):
```yaml
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  name: test-hubot
spec:
  releaseName: test-hubot
  helmVersion: v3
  chart:
    repository: https://kubernetes-charts.storage.googleapis.com/
    name: hubot
    version: 1.0.1
```
2. Run `watch helm ls` and notice that a new revision is created roughly every 3 minutes

**Expected behavior**
The release is not upgraded unless there is an actual change in the chart, from a bumped version or updated values.

**Additional context**
- Helm-operator chart v0.4.0
- Helm-operator docker image v1.0.0-rc6

hiddeco:
Fixes #194, likely also addresses https://github.com/fluxcd/helm-operator/issues/8#issuecomment-568592870, https://github.com/fluxcd/helm-operator/issues/8#issuecomment-569815387

### DIFF
--- a/pkg/helm/helm.go
+++ b/pkg/helm/helm.go
@@ -4,8 +4,8 @@ import "sync"
 
 // Client is the generic interface for Helm (v2 and v3) clients.
 type Client interface {
+	Get(releaseName string, opts GetOptions) (*Release, error)
 	UpgradeFromPath(chartPath string, releaseName string, values []byte, opts UpgradeOptions) (*Release, error)
-	Status(releaseName string, opts StatusOptions) (*Release, error)
 	History(releaseName string, opts HistoryOptions) ([]*Release, error)
 	Rollback(releaseName string, opts RollbackOptions) (*Release, error)
 	DependencyUpdate(chartPath string) error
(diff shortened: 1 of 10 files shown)


stefanprodan (approved):
LGTM

PR #210, opened 2020-01-16
status: retry status and condition updates on conflict

hiddeco:
Spent some time looking into a way to reduce the amount of status pushes in the `release` package but have not settled on an approach (yet).


stefanprodan (approved):
LGTM

PR #213, opened 2020-01-16
release: record rolled back chart revision

hiddeco:
By comparing this recorded revision we are able to detect if we have
already performed an upgrade for a given revision in case the chart
cache has been cleared.


stefanprodan (approved):
LGTM

PR #219, opened 2020-01-17
Allow selecting configmaps and secrets from other namespaces in valuesFrom

### LINKED ISSUE #120: Question: How to use common environment variables or shared config-maps across chart
Not sure if this will become a feature request, or if this is possible with the version we have, or will be possible with the new Kustomize feature. We have started using Flux in development and production clusters. For development we create new clusters on a, almost, weekly basis. All development clusters points to the same branch and config in the repo. Before we installed flux we installed the components to the cluster through a custom script, but we want to move most of this to Flux config. We have started to see that it would be nice to have a mechanism to share config across components. However we struggle when having components in different namespaces and/or where the config for each component is named differently in the different charts. An example of such a config is the **clusterName**. For the first component we flux'ified, our own controller which runs in default namespace, we created a configmap which feeds the values into the Helm-release. For that component the **clusterName** is a top level value in the values file. The second component was Velero, which needed to be inside a velero namespace. The **clusterName** went into the `configuration -> backupStorageLocation -> bucket`. The third component, kubed, would need to has its **clusterName** config in `config -> clusterName`, the fourth, external-dns, gets **clusterName** into `txtOwnerId`. The way we know how to solve this would be to create a config-map for each namespace where there is a component which should be managed by flux, which needs dynamic variables from outside. For components living in the same namespace, to have a shared configmap where the clustername is duplicated to different sections of that configmap.

What would be nice to be able to do was to either:
- Tell flux about cluster global environment variables, through i.e. ``extraEnvs`` that we could set when we install to that specific cluster, and refer to that variable in the Helmrelease when setting the specific value
- Or to have a `configMapKeyRef` to a shared configmap in the cluster. I.e. having a configmap in the default namespace holding all, even if values are duplicated to different sections, cluster global environment variables, and allow flux to read from that configmap

carlpett:
Fixes #120

This PR adds the ability for ConfigMaps and Secrets in other namespaces than the HelmRelease to be referenced.
To accomplish this, the `ConfigMapKeyRef` and `SecretKeyRef` have new types after discussions on Slack. I believe the change is backwards compatible, but please have an extra look on that in reviewing.


pkg/apis/helm.fluxcd.io/v1/types.go
```diff
@@ -89,6 +89,23 @@ type ValuesFromSource struct {
 	ChartFileRef *ChartFileSelector `json:"chartFileRef,omitempty"`
 }
 
+type NamespacedConfigMapKeySelector struct {
+	corev1.LocalObjectReference `json:",inline" protobuf:"bytes,1,opt,name=localObjectReference"`
```
stefanprodan:
```suggestion
	Name string `json:"name"`
```

Please remove all protobuf annotations, we don't use those in CRDs

pkg/apis/helm.fluxcd.io/v1/types.go
```diff
@@ -89,6 +89,23 @@ type ValuesFromSource struct {
...
+	Key                         string `json:"key" protobuf:"bytes,2,opt,name=key"`
+	// +optional
+	Optional *bool `json:"optional,omitempty" protobuf:"varint,3,opt,name=optional"`
+	// +optional
+	Namespace string `json:"optional,omitempty" protobuf:"bytes,4,opt,name=namespace"`
+}
+type NamespacedSecretKeySelector struct {
+	corev1.LocalObjectReference `json:",inline" protobuf:"bytes,1,opt,name=localObjectReference"`
```
stefanprodan:
```suggestion
	Name string `json:"name"`
```
Please remove all protobuf annotations, we don't use those in CRDs

stefanprodan (changes requested):
The namespace should be added to docs as an optional field https://github.com/fluxcd/helm-operator/blob/master/docs/references/helmrelease-custom-resource.md#secrets

carlpett:
@stefanprodan Thanks! Pushed updates.

pkg/apis/helm.fluxcd.io/v1/types.go
```diff
@@ -89,6 +89,23 @@ type ValuesFromSource struct {
 	ChartFileRef *ChartFileSelector `json:"chartFileRef,omitempty"`
 }
 
+type NamespacedConfigMapKeySelector struct {
+	corev1.LocalObjectReference `json:",inline"`
```
stefanprodan: I don't think we should embed the `LocalObjectReference` since it could change in the future and it would conflict with the `Key` field. Also the `LocalObjectReference` has the `Name` field marked as optional but this would break Helm operator, a secret ref can't be resolved if the name is not specified. 

Instead of embedding  `LocalObjectReference` we could use
```suggestion
	Name string `json:"name"`
```

stefanprodan (changes requested):
The namespace field should be added to the CRD validation spec:
- https://github.com/fluxcd/helm-operator/blob/master/deploy/flux-helm-release-crd.yaml#L104-L129
- https://github.com/fluxcd/helm-operator/blob/master/chart/helm-operator/templates/crd.yaml#L112-L137

pkg/apis/helm.fluxcd.io/v1/types.go
carlpett: I pretty much lifted the object definition from the earlier `corev1.ConfigMapKeySelector` and tried to modify it as little as possible. I don't recall the specifics of backwards compatibility regarding loading previous versions from storage when changing the types, but if that isn't an issue, then yes, definitely better to change it.

stefanprodan (changes requested):
The namespace filed has to be added to the CRD in chart/helm-operator/templates and in https://github.com/fluxcd/helm-operator/blob/master/pkg/install/templates/flux-helm-release-crd.yaml.tmpl. You also need to regenerate the install embed files with `make generate-deploy`.

carlpett:
Those sure are a lot of places to remember to change :smile: 
Just to verify - the change itself looks okay?

stefanprodan:
The PR looks ok to me, before removing WIP please rebase with master and squash/reword the last commits.

carlpett:
Rebased and squashed

stefanprodan (approved):
LGTM

Thanks @carlpett 🏅

carlpett:
Did a final read-through myself and found a copy/paste error :man_facepalming: Pushing a fix in a minute

carlpett:
Pushed a configmap->secret typo fix in pkg/install/templates/flux-helm-release-crd.yaml.tmpl + `make generate-deploy`.

carlpett:
@stefanprodan Looks like I managed to break a e2e test somehow with my spelling correction... I don't really see how, though, unless there known flakes there?

stefanprodan:
I've rerun the tests all good It seems that the v3 rollback tests are flaky.

hiddeco (approved):

PR #223, opened 2020-01-20
Reject git source if URL and path are missing

stefanprodan:
Helm operator choses to use a Git source when a chart comes from a Helm repository if `ref` is specified. This PR fixes the Git source detection by validating the git URL and path.

HR release spec that is mistaken for a Git source:
```yaml
spec:
  chart:
    name: expressjs-k8s
    ref: master
    repository: https://alexellis.github.io/expressjs-k8s/
    version: 0.1.1
```


hiddeco (approved):
LGTM

Issue #108, opened 2019-11-17
Flux Helm Operator freezes....


derrickburns:
See: 

https://github.com/solo-io/gloo/issues/1730

derrickburns:
And another instance:
```
│ ts=2019-11-22T19:15:19.352752162Z caller=chartsync.go:207 component=chartsync warning="could not get revision for ref while checking for changes" resource=qa1:helmrelease/tidepool repo=gi │
│ t@github.com:tidepool-org/development ref=redirect err="fatal: bad revision 'redirect', full output:\n fatal: bad revision 'redirect'\n"
```
Then I pushed the branch.

Then:

```
ts=2019-11-22T19:25:20.372269259Z caller=chartsync.go:251 component=chartsync info="enqueing release upgrade due to change in git chart source" resource=qa1:helmrelease/tidepool           │
│ ts=2019-11-22T19:25:28.422815897Z caller=release.go:184 component=release info="processing release tidepool-qa1 (as 45ed6972-ed7e-11e9-98b7-02a97237eade)" action=CREATE options="{DryRun:t │
│ rue ReuseName:false}" timeout=300s
```

with no progress!

derrickburns:
This was caused by a templating error in the helm chart that it was reading.

```
      - prefix: '/'redirectAction:
```

derrickburns:
@hiddeco 
I am still getting freezes.

```
Last log entry:
ts=2019-12-12T18:52:07.368127641Z caller=release.go:184 component=release info="processing release tidepool-production (as d2b038e8-f055-11e9-bf99-0e6fbf997a04)" action=CREATE options="{DryRun:true ReuseName:false}" timeout=300s
```

current time: 17 minutes later

In this case, there was another templating error.

```
blah
  ---
apiversion: blah
```
The extra spaces before the three hyphens is an error.

At minimum, I suggest that the helm operator should fail a liveness test and be automatically restarted.  It does not.

hiddeco:
@derrickburns does this issue still exist for one of the latest `1.0.0-rcX` releases? Much of the `chartsync` spaghetti code has been refactored so I can imagine this got (indirectly) resolved.

derrickburns:
@hiddeco i have not observed problem recently. 

Thx

hiddeco:
@derrickburns awesome, thanks for confirming!

derrickburns:
@hiddeco  Please reopen.  It just froze again:

Here is the log:

```
[...]

PR #225, opened 2020-01-20
helm: lock and index fetch improvements

### LINKED ISSUE #133: Dry run installs are processed serially
Hello there 👋 

**Describe the bug**
I have noticed that even though I have set the numbers of workers to 4, the Helm operator seems to dry-run install releases serially. I verified using a chart that takes about 2 minutes to be rendered and the queue is plateauing as such:

Capture d’écran 2019-12-04 à 13 59 54

As you can see above, the Helm operator performs a dry-run install of all (about 60) releases every `chartsSyncInterval` (10min in my case) but it seems they're done in a serial fashion.

**To Reproduce**
- Set the numbers of Helm operator workers to >1
- Have a bunch of HelmReleases in cluster
- Create a dummy HelmRelease based on dummy chart that takes time to install, e.g. with the following template:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: dummy
  {{ range $i, $e := until 500000000}}{{end -}}
  {{ range $i, $e := until 500000000}}{{end -}}
  {{ range $i, $e := until 500000000}}{{end -}}
  {{ range $i, $e := until 500000000}}{{end -}}
```

**Expected behavior**
The Helm operator processes dry-run installs on its workers in parallel.

**Logs**
Not applicable.

**Additional context**
- Helm operator version: 1.0.0-rc4
- Kubernetes version: v1.14.8-gke.12

hiddeco:
Fixes #133 

This likely also addresses part of #108, but should not be considered a fix as it is a postponement of execution.

### DIFF
--- a/pkg/helm/v2/dependency.go
+++ b/pkg/helm/v2/dependency.go
@@ -7,9 +7,6 @@ import (
 )
 
 func (h *HelmV2) DependencyUpdate(chartPath string) error {
-	repositoryConfigLock.RLock()
-	defer repositoryConfigLock.RUnlock()
-
 	out := helm.NewLogWriter(h.logger)
 	man := downloader.Manager{
 		Out:       out,
--- a/pkg/helm/v2/pull.go
+++ b/pkg/helm/v2/pull.go
@@ -1,6 +1,10 @@
 package v2
 
 import (
+	"os"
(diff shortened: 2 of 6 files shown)


stefanprodan (approved):
LGTM

PR #229, opened 2020-01-21
chart: Allow pod annotations, labels and account annotations to be set

stefanprodan:
Allow custom configuration for:
- pod annotations
- pod labels
- service account annotations

Set replica to 1 as multiple replicas are not supported.

### DIFF
--- a/chart/helm-operator/templates/deployment.yaml
+++ b/chart/helm-operator/templates/deployment.yaml
@@ -8,7 +8,7 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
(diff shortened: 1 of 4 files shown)


chart/helm-operator/templates/deployment.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
-  replicas: {{ .Values.replicaCount }}
+  replicas: 1
```
hiddeco, line 11: Heh, good catch :fish:

hiddeco (approved):
LGTM :100:

PR #230, opened 2020-01-21
Update Flux module to v1.17.2-0.20200121140732-3903cf8e71c3

### LINKED ISSUE #217: Broken dependency reference
**Describe the bug**

The version of the [github.com/fluxcd/flux](https://github.com/fluxcd/flux) dependency that is referenced by this project corresponds to a commit that no longer exists in the upstream, thus making it impossible to build for people not using the Go Module proxy.

**To Reproduce**

On Go 1.13.x force-disable the use of the module proxy by setting `go env -w GOPROXY=direct` (which can later be reverted by running `go env -w GOPROXY=`).

Ensure your module cache does not contain the version of the `github.com/fluxcd/flux` currently required by this project. Instead of cleaning the entire mod-cache via `go clean -modcache` the hacky way of doing this is via
```shell
rm -rf "$(go env GOPATH)"/pkg/mod/{,cache/download/}github.com/fluxcd/flux/@v/v1.17.1-0.20200109130101-ca3e8f451dd0.*"
```

**Additional context**

This was discovered during the investigation of golang/go#36624.

stefanprodan:
Fix: #217

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.13
 
 require (
 	github.com/Masterminds/semver v1.5.0 // indirect
-	github.com/fluxcd/flux v1.17.1-0.20200109130101-ca3e8f451dd0
+	github.com/fluxcd/flux v1.17.2-0.20200121140732-3903cf8e71c3
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-kit/kit v0.9.0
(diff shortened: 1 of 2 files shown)


hiddeco (approved):
Bit anxious about tying it to `master`, but what needs to be done needs to be done (and it is commit-versioned).

PR #234, opened 2020-01-21
Build: make pkg/install a Go module to reduce its dependencies

hiddeco:


stefanprodan:
This partially solves #238

stefanprodan (approved):
LGTM

PR #235, opened 2020-01-21
release: allow MaxHistory to be overriden for a HelmRelease

### LINKED ISSUE #224: Helmrelease spec should contain --history-max for v3 releases
**Describe the feature**
A clear and concise description of the feature.
Currently is it not possible to set the number of helmv3 revision, which defaults to 10.
The --history-max int flag provide this capability  in the helm upgrade command.
**What would the new user story look like?**
New spec historyMax in the helmrelease file should set the number of helm revisions
**Expected behavior**
A clear and concise description of what you expect to happen.
If set the specified number of historyMax revision files should be keep in the cluster

stefansedich:
Currently the --history-max=10 was hard-coded and not user configurable, this adds `maxHistory` to the HelmRelease CRD to allow a user to override this if desired.

@hiddeco hoping I got all the required bits this is my first CRD related PR let me know what I am missing.

Closes #224


pkg/release/release.go
```diff
@@ -195,7 +193,14 @@ func (r *Release) Sync(client helm.Client, hr *v1.HelmRelease) (rHr *v1.HelmRele
 	// technically return an error.
 	v, _ := composedValues.YAML()
 
-	var performRollback bool
+	var (
+		performRollback bool
+		maxHistory      int = 10
```
hiddeco: Instead of setting this here I would simply move the default to the `HelmRelease` itself, an example of how this is done [can be found here](https://github.com/fluxcd/helm-operator/blob/master/pkg/apis/helm.fluxcd.io/v1/types.go#L217).

hiddeco (commented):
Can you also add the CRD change in the Helm chart to make sure we do not forget about this later on?

pkg/release/release.go
stefansedich: Done!

stefansedich:
@hiddeco added to CRD I assume this was manual as I had done?

hiddeco (approved):
Splendid! :1st_place_medal:

test/e2e/25_max_history.bats
```diff
@@ -0,0 +1,70 @@
...
+  for i in {1..15}
+  do
+    # Apply a patch to initiate an upgrade
+    kubectl patch -f "$FIXTURES_DIR/releases/helm-repository.yaml" --type='json' -p="[{\"op\": \"replace\", \"path\": \"/spec/values/someField\", \"value\":\"$i\"}]" >&3
+    poll_until_equals 'podinfo-helm-repository HelmRelease' "$(($i+1))" "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-helm-repository -o 'custom-columns=status:status.observedGeneration' --no-headers"
+  done
+
+  # Check count of revisions is <= 10
+  count=$(k get secret -n $DEMO_NAMESPACE --output name | grep 'sh.helm.release.v1.podinfo-helm-repository' | wc -l)
+  [ "$count" -le 10 ]
```
hiddeco: I think it is wiser to not use our human knowledge here about the configured storage mechanism but instead `exec` into the running container and use the `helm3` binary there to get this information. This would make the end-to-end tests still succeed if for some reason the storage mechanism is configured differently.

I also think the stricter `[ "$count" -eq 10 ]` can be used, given we use `poll_until_equals` in the `for` loop and there should thus be no divergence due to e.g. the operator not being ready with processing.

test/e2e/25_max_history.bats
```diff
@@ -0,0 +1,70 @@
...
+
+  # Patch HelmRelease to set maxHistory to 5
+  kubectl patch -f "$FIXTURES_DIR/releases/helm-repository.yaml" --type='json' -p='[{"op": "replace", "path": "/spec/maxHistory", "value":5}]' >&3
+  poll_until_equals 'podinfo-helm-repository HelmRelease' '2' "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-helm-repository -o 'custom-columns=status:status.observedGeneration' --no-headers"
+
+  for i in {1..8}
+  do
+    # Apply a patch to initiate an upgrade
+    kubectl patch -f "$FIXTURES_DIR/releases/helm-repository.yaml" --type='json' -p="[{\"op\": \"replace\", \"path\": \"/spec/values/someField\", \"value\":\"$i\"}]" >&3
+    poll_until_equals 'podinfo-helm-repository HelmRelease' "$(($i+2))" "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-helm-repository -o 'custom-columns=status:status.observedGeneration' --no-headers"
```
hiddeco, line 54: What is the reason for `$(($i+2))` here instead of `$(($i+1))`?

test/e2e/25_max_history.bats
```diff
@@ -0,0 +1,70 @@
...
+  done
+
+  # Check count of revisions is <= 5
+  count=$(k get secret -n $DEMO_NAMESPACE --output name | grep 'sh.helm.release.v1.podinfo-helm-repository' | wc -l)
```
hiddeco: Same comment as the other test above.

test/e2e/25_max_history.bats
stefansedich: Reason being I apply the patch above to set the maxHistory so to start we are at revision 2 before we begin this loop.

hiddeco: I think I wasn't fully awake yet this morning. Makes totally sense now.

stefansedich: Is there a better way to wait for the rollout? it was all I could think of that was reliable.

test/e2e/25_max_history.bats
```diff
@@ -0,0 +1,70 @@
...
+
+  for i in {1..8}
+  do
+    # Apply a patch to initiate an upgrade
+    kubectl patch -f "$FIXTURES_DIR/releases/helm-repository.yaml" --type='json' -p="[{\"op\": \"replace\", \"path\": \"/spec/values/someField\", \"value\":\"$i\"}]" >&3
+    poll_until_equals 'podinfo-helm-repository HelmRelease' "$(($i+2))" "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-helm-repository -o 'custom-columns=status:status.observedGeneration' --no-headers"
+  done
+
+  # Check count of revisions is <= 5
+  count=$(kubectl exec -n "$E2E_NAMESPACE" deploy/helm-operator -- helm3 history podinfo-helm-repository --namespace "$DEMO_NAMESPACE" --skip-headers | tail -n +2 | wc -l)
```
stefansedich, line 58: Could not find a way to not output headers like you can with kubectl in table mode so skipping the first line.

test/e2e/25_max_history.bats
```diff
@@ -0,0 +1,70 @@
...
+
+  for i in {1..15}
+  do
+    # Apply a patch to initiate an upgrade
+    kubectl patch -f "$FIXTURES_DIR/releases/helm-repository.yaml" --type='json' -p="[{\"op\": \"replace\", \"path\": \"/spec/values/someField\", \"value\":\"$i\"}]" >&3
+    poll_until_equals 'podinfo-helm-repository HelmRelease' "$(($i+1))" "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-helm-repository -o 'custom-columns=status:status.observedGeneration' --no-headers"
+  done
+
+  # Check count of revisions is <= 10
+  count=$(kubectl exec -n "$E2E_NAMESPACE" deploy/helm-operator -- helm3 history podinfo-helm-repository --namespace "$DEMO_NAMESPACE" --skip-headers | tail -n +2 | wc -l)
```
stefansedich, line 33: Could not find a way to not output headers like you can with kubectl in table mode so skipping the first line.

test/e2e/25_max_history.bats
hiddeco: Looks like the `-q` option available for other commands is not available for the `history` command Helm has.

hiddeco (approved):
LGTM, thanks for adding the end-to-end tests. :star:

PR #242, opened 2020-01-24
helm: do not include non-template files in chart data

### LINKED ISSUE #194: Charts with dependencies are upgraded every cycle

### LINKED ISSUE #218: repeated upgrades with helm operator even though there is no change
**Describe the bug**
I have 3 helm releases , out of one is upgrading every min even though the there is no change made  

**Logs**

Screen Shot 2020-01-17 at 6 21 04 PM

**Additional context**
- Helm operator version: 
   ```docker.io/fluxcd/helm-operator:1.0.0-rc7```   
- Git provider:
   bitbucket
- Container registry provider:
   acr

Kind of same issue https://github.com/fluxcd/helm-operator/issues/8#issuecomment-574389970

Is there any workaround to solve this kind of issue ?

hiddeco:
As the non-template files include the `requirements.lock` file,
which in case of a chart with dependencies differs every time
a dry-run happens due to the generation timestamp it includes,
resulting in spurious upgrades.

Other files included there are non-trivial for the detection of
changes as their effects are either captured in other parts of
the chart metadata, or do not matter at all for the release
itself, like the contents of a `README.md`.

Fixes #218

### DIFF
--- a/pkg/helm/release.go
+++ b/pkg/helm/release.go
@@ -48,7 +48,6 @@ type Chart struct {
 	Version      string
 	AppVersion   string
 	Values       Values
-	Files        []*File
 	Templates    []*File
 }
 
--- a/pkg/helm/v2/release.go
+++ b/pkg/helm/v2/release.go
@@ -42,7 +42,6 @@ func chartToGenericChart(c *chart.Chart) *helm.Chart {
 		Version:      c.Metadata.Version,
 		AppVersion:   c.Metadata.AppVersion,
 		Values:       valuesToGenericValues(c.Values),
-		Files:        filesToGenericFiles(c.Files),
 		Templates:    templatesToGenericFiles(c.Templates),
 	}
 }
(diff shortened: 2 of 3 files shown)


stefanprodan (approved):
LGTM

PR #246, opened 2020-01-26
install: Add Helm versions and extra args to install options

stefanprodan:
Allow eksctl and other consumers of the install package to configure the Helm versions or other cmd args.


tools.go
```diff
@@ -0,0 +1,8 @@
+// +build tools
+
+// This file just exists to ensure we download the tools we need for building
+// See https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
+
+package helm_operator
+
+import "github.com/fluxcd/helm-operator/pkg/install"
```
hiddeco: Newline

hiddeco (approved):
Thanks Stefan 👏

tools.go
stefanprodan: I went with the import format used in Flux

PR #251, opened 2020-01-27
Fix default value of helm.version

cbenjemaa:
The default value is `v2`


stefanprodan:
The default value is `v2,v3` see https://github.com/fluxcd/helm-operator/blob/master/chart/helm-operator/values.yaml#L35 and https://github.com/fluxcd/helm-operator/blob/master/cmd/helm-operator/main.go#L127

PR #252, opened 2020-01-27
release: support retrying rollbacks

### LINKED ISSUE #241: Retry upgrade after rollback without change to resource
@hiddeco 

I think that the guard on whether to retry a release after a rollback is too restrictive:

https://github.com/fluxcd/helm-operator/blob/83c8f80bcde6ab5a38553349c5d75a995be6ee1c/pkg/release/release.go#L326

For example, suppose an install falls because of a timeout or because a dependencies are not satisfied such as a ServiceAccount being missing. None of these will appear as changes in values. 

More generally, a release may fail and be rolled back because of externalities and these externalities may be transient.

I believe that I have encountered precisely these failures.

Is my concern warranted?

hiddeco:
This commit adds support for retrying rollbacks using `.spec.rollback.retry` until the configured `.spec.rolback.maxRetries` count is reached (defaults to 5, setting it to 0 means infinite).

This makes it possible to for example recover from transient errors that were the cause of the rollback, and not a persistent chart or values failure. The reset of the counter happens after a successful
upgrade or a change (e.g. chart change, `.spec` change).

Fixes #241


stefanprodan (approved):
LGTM 👍

PR #256, opened 2020-01-29
release: disable atomic flag for initial install

### LINKED ISSUE #255: Add property to disable atomic installation
In the early versions of the operator, the installation would not wait for the chart to be ready by default. That could be changed using the rollback property in the HelmRelease.

After [this change](https://github.com/fluxcd/helm-operator/commit/30156d40e1cb4de27f6c945de3af1a72e667e9ad), the default behavior is that, for the installation only, the operator will wait (atomic flag) until the chart is ready. The problem is that there is no way to disable that.

Given the following scenario:

* Repo with several HelmReleases (let's say 15)
* Some charts depend on others.
* The cluster is bootstrapped and automatically installed Flux.
* The Helm operator will start to install each release in no specific order taking 2 at the time.

This will lead to a case in which each installation will fail (for timeout) until the dependencies are meet. At the end, the operator will install everything, but it will take much longer than before.

It would be nice to have the possibility to disable the atomic behavior during the chart installation.

Thanks!

hiddeco:
As it does not play nicely with some bootstrapping scenarios, enabling the (waiting) behaviour is still possible by setting `.spec.wait` to `true`.

Fixes #255

### DIFF
--- a/pkg/release/release.go
+++ b/pkg/release/release.go
@@ -202,20 +202,29 @@ func (r *Release) Sync(client helm.Client, hr *v1.HelmRelease) (rHr *v1.HelmRele
 		Force:       hr.Spec.ForceUpgrade,
 		ResetValues: hr.Spec.ResetValues,
 		MaxHistory:  hr.GetMaxHistory(),
-		// We only set this during installation to delete a failed
-		// release, but not during upgrades, as we ourselves want
-		// to be in control of rollbacks.
-		Atomic: curRel == nil,
-		Wait:   hr.Spec.Wait || hr.Spec.Rollback.Enable,
+		Wait:        hr.Spec.Wait || (curRel != nil && hr.Spec.Rollback.Enable),
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

marshallford:
Would it be reasonable to set `.spec.wait` to `true` by default to prevent a breaking change?

hiddeco:
@marshallford no, as the behaviour should not have been introduced to start with and the breaking change being made was actually _using_ the `--atomic` flag (see linked issue).

marshallford:
I guess there was two breaking changes then, because I was relying on the `--atomic` being present by default.

hiddeco:
It depends on the moment you started using the operator, but note that:

1. part of the `--atomic` behaviour is still present (as a failed installation still results in the removal of the release)
2. the atomic behaviour only happened on chart installations, given those are already installed in your cluster, this change should have no effect on those existing releases

Issue #215, opened 2020-01-17
Proper way to migrate from Helm Operator v2 only to Helm Operator v3 only


derrickburns:
This is not a bug report as much as it is a plea for help.

How does one convert from using Helm Operator with Helm v2 only support to
Helm Operator with Helm v3 only support (ignoring installation).

I tried :
1) scale down flux and helm operator  (helm v2 only) to zero
2) convert helm using `helm 2to3 convert ...`
3) delete flux namespace
4) install new flux and helm operator (helm v3 only)

And I also tried this without step (2).

The  second approach was a disaster.  Helm Operator refused to overwrite existing resources. So I had to manually remove them first.

Suggestions?

hiddeco:
I would expect the 4 steps you have documented to work as long as the converted releases are written to the `secret` storage. We actually want to make it possible to automagically migrate resources using `2to3` in one of the future releases.

derrickburns:
One important detail: the converted secrets MUST be placed in the same namespace as the helmrelease.

hiddeco:
@derrickburns will keep that in mind while utilizing it, thanks.

---

@stefanprodan any ideas on what the most straight forward approach would be for users to tell the operator that they want to migrate a release from `v2` to `v3`?

derrickburns:
The namespace issue that i mentioned was fixed in a newer version of the plugin.

mcfearsome:
Very interested in this, I have about 60 clusters to migrate at some point without disturbing the cluster.

derrickburns:
```
# install helm3 see  https://helm.sh/docs/intro/install/
# then install 2to3 plugin
helm plugin install https://github.com/helm/helm-2to3
```

@mcfearsome I had success with this script after installing helm3 and the `2to3` plugin:

```
    export FLUX_FORWARD_NAMESPACE=flux
    kubectl scale deployment --replicas=0 -n ${FLUX_FORWARD_NAMESPACE} flux-helm-operator
    kubectl scale deployment --replicas=0 -n ${FLUX_FORWARD_NAMESPACE} flux
    kubectl get configmap -n ${FLUX_FORWARD_NAMESPACE} -l "OWNER=TILLER" | awk '{print $1}' | grep -v NAME | cut -d '.' -f1 | uniq | xargs -n1 helm 2to3 -t ${FLUX_FORWARD_NAMESPACE} convert
    kubectl delete ns ${FLUX_FORWARD_NAMESPACE}
[...]

PR #264, opened 2020-01-31
helm/v3: add v2 -> v3 release converter

hiddeco:
Addresses #215


saada:
As discussed with @hiddeco , I'll be picking up this PR

PR #269, opened 2020-02-05
Make sure we test all the local modules

2opremio:
Followup to #234

Equivalent to Flux's https://github.com/fluxcd/flux/pull/2822


2opremio:
Flakey test?

hiddeco (approved):

PR #270, opened 2020-02-05
Generate Helm chart and `deploy/` CRD(s) using `controller-gen`

hiddeco:


hack/tools/go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/fluxcd/helm-operator/hack/tools
 go 1.13
 
 require (
-	github.com/spf13/pflag v1.0.5 // indirect
 	golang.org/x/tools v0.0.0-20190927052746-69890759d905 // indirect
-	k8s.io/code-generator v0.16.5-beta.1 // kubernetes-1.16.2
+	k8s.io/code-generator v0.17.0 // kubernetes-1.16.2
```
stefanprodan: I think the generated code and the controller should use the same Kubernetes version, otherwise we risk to have an incompatible clientset.

hack/update/generate-crds.sh
```diff
@@ -0,0 +1,70 @@
...
+"${REPO_ROOT}/hack/go_container.sh" go build -o /out/controller-gen sigs.k8s.io/controller-tools/cmd/controller-gen
+# go back to the root
+cd "${REPO_ROOT}"
+
+# turn off module mode before running the generators
+# https://github.com/kubernetes/code-generator/issues/69
+# we also need to populate vendor
+hack/go_container.sh go mod tidy
+hack/go_container.sh go mod vendor
+export GO111MODULE="off"
```
stefanprodan, line 37: Do we need this for Kubernetes > 1.16?

hiddeco: No idea, and worth and investigation, but probably better to do this in a separate PR.

deploy/helm-operator-crds.yaml
```diff
@@ -0,0 +1,355 @@
...
+                type: object
+                required:
+                - name
+                properties:
+                  name:
+                    type: string
+            values:
+              description: Values holds the values for this Helm release.
+              type: object
+              x-kubernetes-preserve-unknown-fields: true
```
stefanprodan: This will break compatibility with Kubernetes < 1.17, people will have to disable kubectl validation to apply it.

hiddeco: It works for Kubernetes 1.15 and above, but fair point for our beloved GKE/EKS users. I have removed the flag putting it there.

stefanprodan (approved):
LGTM

hiddeco:
@stefanprodan going to merge this and do the renaming of the manifests and stuff in a separate PR, as due to the discussions this is going to be a bit more than just renaming the files.

PR #287, opened 2020-02-11
chart: use Helm v3 CRD approach with backwards compatibility for v2

stefansedich:
Move chart CRDs into crds/ folder for helm3 CRD support and change so that `make generate-deploy` will copy the generated CRD to the chart removing the manual step previously required.


Makefile
```diff
@@ -138,6 +138,7 @@ cache/bats-core-$(BATS_COMMIT).tar.gz:
 
 generate-deploy: pkg/install/generated_templates.gogen.go
 	cd deploy && go run ../pkg/install/generate.go deploy
+	cp ./deploy/flux-helm-release-crd.yaml ./chart/helm-operator/crds/helmrelease.yaml
```
stefansedich, line 141: Removes the need to manually modify the chart CRD when changes are made,

hiddeco (approved):
Once again saving me tons of work, thanks a lot :star2:

PR #289, opened 2020-02-12
build: adding generate-codegen step to Makefile

stefansedich:
After having had to play the generate dance a few times now, feels like adding a `make generate-go` step that will do the Go generation will make life easier and with the addition of a `make generate` step to do all the generating at once.

This will also simplify things as I write up the Getting started developing docs (coming soon).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,11 @@ cache/bats-core-$(BATS_COMMIT).tar.gz:
 	# Use 2opremio's fork until https://github.com/bats-core/bats-core/pull/255 is merged
 	curl --fail -L -o $@ https://github.com/2opremio/bats-core/archive/$(BATS_COMMIT).tar.gz
 
+generate: generate-codegen generate-deploy
+
+generate-codegen:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -136,6 +136,11 @@ cache/bats-core-$(BATS_COMMIT).tar.gz:
 	# Use 2opremio's fork until https://github.com/bats-core/bats-core/pull/255 is merged
 	curl --fail -L -o $@ https://github.com/2opremio/bats-core/archive/$(BATS_COMMIT).tar.gz
 
+generate: generate-go generate-deploy
+
+generate-go:
```
hiddeco: I made this same addition in #270 but called it `generate-codegen` there, which is a bit more descriptive to me personally.

stefansedich: As usual naming stuff is hard :smile: updated!

hiddeco (approved):
LGTM

stefansedich:
Rebased

PR #290, opened 2020-02-12
build: do not create kind clusters in parallel

stefansedich:
Noticing the test flakes and saw the discussion in https://github.com/kubernetes-sigs/kind/issues/1288 changing to 2 parallel jobs to see if that has any effect on stability of the e2e suite in CI, @hiddeco any objection to trying this out?

I also factored the e2e tests out into a CircleCI command which IMO cleans things up a little but required an update to the 2.1 build (which should not break anything else).

### DIFF
--- a/test/e2e/lib/install.bash
+++ b/test/e2e/lib/install.bash
@@ -7,16 +7,18 @@ source "${E2E_DIR}/lib/template.bash"
 
 function install_tiller() {
   if ! helm2 version > /dev/null 2>&1; then # only if helm isn't already installed
-    kubectl --namespace kube-system create sa tiller
-    kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
-    helm2 init --service-account tiller --upgrade --wait
+    kubectl --namespace "$E2E_NAMESPACE" create sa tiller
+    kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount="$E2E_NAMESPACE":tiller
+    helm2 init --tiller-namespace "$E2E_NAMESPACE" --service-account tiller --upgrade --wait
   fi
 }
 
 function uninstall_tiller() {
-  helm2 reset --force
+  # Note: helm reset --force will delete the Tiller
+  # instance but will not delete release history.
+  helm2 reset --tiller-namespace "$E2E_NAMESPACE" --force
   kubectl delete clusterrolebinding tiller-cluster-rule
-  kubectl --namespace kube-system delete sa tiller
+  kubectl --namespace "$E2E_NAMESPACE" delete sa tiller
 }
 
 function install_helm_operator_with_helm() {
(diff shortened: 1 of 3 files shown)


stefansedich:
Looks like `6 When rollback.maxRetries is set to 1,  upgrade is only retried once` is giving me grief but looking at previous builds looks like that test has been known to do that.

hiddeco:
@stefansedich looks like changing this to 2 clusters has made the flakiness of the test persistent, while this is great to debug what goes wrong, it doesn't help to reduce the flakiness of the cluster bootstrapping.

hiddeco:
For the (public) record: persistent failure happens due to Tiller not being properly cleaned up in the teardown, as `helm reset --force` deletes the instance but keeps the releases.

Two options:
1. Remove all releases before nuking Tiller
2. Install Tiller into the same namespace as the operator is in, this would require a bit more wiring but would give the advantage of testing the Tiller configuration flags while also making the teardown itself easier as the namespace can just be nuked.

stefansedich:
Have gone for option #2, pushed and :crossed_fingers: for a solid CI experience.

test/e2e/lib/install.bash
```diff
@@ -7,16 +7,16 @@ source "${E2E_DIR}/lib/template.bash"
...
+    helm2 init --tiller-namespace "$E2E_NAMESPACE" --service-account tiller --upgrade --wait
   fi
 }
 
 function uninstall_tiller() {
-  helm2 reset --force
+  helm2 reset --tiller-namespace "$E2E_NAMESPACE" --force
   kubectl delete clusterrolebinding tiller-cluster-rule
-  kubectl --namespace kube-system delete sa tiller
+  kubectl --namespace "$E2E_NAMESPACE" delete sa tiller
```
stefansedich, line 21: I guess this isn't really needed as we nuke the namespace but it does not hurt.

hiddeco: This function is not aware of how it is utilized and should simply undo what its `install` companion creates, so this is perfect.

test/e2e/lib/install.bash
```diff
@@ -7,16 +7,16 @@ source "${E2E_DIR}/lib/template.bash"
...
-    helm2 init --service-account tiller --upgrade --wait
+    kubectl --namespace "$E2E_NAMESPACE" create sa tiller
+    kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount="$E2E_NAMESPACE":tiller
+    helm2 init --tiller-namespace "$E2E_NAMESPACE" --service-account tiller --upgrade --wait
   fi
 }
 
 function uninstall_tiller() {
-  helm2 reset --force
+  helm2 reset --tiller-namespace "$E2E_NAMESPACE" --force
```
hiddeco, line 19: Can you add a note here about the `--force` flag, and the fact that this only resets the Tiller instance and _not_ the release storage.

hiddeco (approved):

PR #291, opened 2020-02-12
helm/v3: add release name to log context for internal Helm logs

### LINKED ISSUE #240: helm/v3: enrich logger with release metadata
```
 2020/01/24 06:02:18 warning: cannot overwrite table with non table for policy (map[])
```

It would be helpful to identify the field that is causing this issue.

stefansedich:
Add release to log context for the logger func that is used to log internal helm logs.

Closes #240

### DIFF
--- a/pkg/helm/v3/get.go
+++ b/pkg/helm/v3/get.go
@@ -9,7 +9,7 @@ import (
 )
 
 func (h *HelmV3) Get(releaseName string, opts helm.GetOptions) (*helm.Release, error) {
-	cfg, err := newActionConfig(h.kubeConfig, h.infoLogFunc, opts.Namespace, "")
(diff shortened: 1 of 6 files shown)


hiddeco (commented):
Think besides adding the `releaseName`, it would probably also good to include the `opts.Namespace` as the same release name can exist in multiple namespaces since v3.

stefansedich:
> Think besides adding the `releaseName`, it would probably also good to include the `opts.Namespace` as the same release name can exist in multiple namespaces since v3.

Makes sense, have added this, just want to do a final test and will take the PR out of draft.

stefansedich:
@hiddeco looks good:

```
ts=2020-02-12T15:03:59.402591551Z caller=helm.go:69 component=helm version=v3 info="creating 8 resource(s)" targetNamespace=demo release=skip-crd
```

PR ready to review.

hiddeco (approved):

PR #292, opened 2020-02-12
pkg: Helm v3.0.3 with global runtime.Scheme patch

### LINKED ISSUE #167: Failing to update chart containing child HelmRelease and nested values
**Describe the bug**

Given the following chart https://github.com/stefansedich/helm-test if one were to apply the outer release.yaml it would install the first time without issue however after an update say bumping the nested image tag we would see the following error: `failed to create patch: merging an object in json but data type is not struct, instead is: map`

Installing directly with helm using `helm upgrade --install podinfo ./chart --values values.yaml` works without issue. And strangely enough if you were to reduce the nesting of `podinfo.some.nested.image` to `podinfo.some.image` helm-operator is able to install and update without issue.

**To Reproduce**

1. kubectl apply -f release.yaml
2. See that it installs fine
3. Change tag in outer release.yaml
4. See the error

**Expected behavior**

Subsequent chart upgrades to work without issue.

**Logs**

```
6 upgrade.go:280] warning: Upgrade "podinfo" failed: failed to create patch: merging an object in json but data type is not struct, instead is: map
ts=2019-12-28T01:35:42.221288399Z caller=release.go:217 component=release release=podinfo targetNamespace=default resource=default:helmrelease/podinfo helmVersion=v3 error="Helm release failed" revision=18651c21dc303955f861fb3e50234afbf9a6620e err="failed to upgrade chart for release [podinfo]: failed to create patch: merging an object in json but data type is not struct, instead is: map"
```

**Additional context**

- Helm operator version: rc5
- Targeted Helm version: v3
- Kubernetes version: 1.14
- Git provider: GitHub
- Container registry provider: N/A

hiddeco:
Closes #167


test/e2e/15_upgrade.bats
```diff
@@ -74,8 +75,43 @@ function setup() {
...
 
   # Assert change is rolled out
-  poll_until_equals 'podinfo-git HelmRelease chart update' "successfully cloned chart revision: $head_hash" "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-git -o jsonpath='{.status.conditions[?(@.type==\"ChartFetched\")].message}'"
-  poll_until_equals 'podinfo-git HelmRelease revision matches' "$head_hash" "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-git -o jsonpath='{.status.revision}'"
+  poll_until_equals 'release chart update' "successfully cloned chart revision: $head_hash" "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-git -o jsonpath='{.status.conditions[?(@.type==\"ChartFetched\")].message}'"
+  poll_until_equals 'revision match' "$head_hash" "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-git -o jsonpath='{.status.revision}'"
+}
+
+@test "When a HelmRelease is nested in a chart, an upgrade does succeed" {
+  # Install chartmuseum
```
hiddeco, line 83: @stefansedich this will probably help you to address the CRD test with a local dummy chart.

stefanprodan (approved):
LGTM

test/e2e/15_upgrade.bats
stefansedich: Great think that will, will update that test and send a PR over.

PR #294, opened 2020-02-12
e2e: attempt to run 8 parallel jobs to speed up e2e tests

stefansedich:
Trying 8 parallel jobs to see the effect that has on build times and if it will work.


stefansedich:
This won't really help as the time is now just burnt upfront creating the clusters, really need to run helm2 and helm3 in parallel if we want some speed gains.

PR #300, opened 2020-02-13
pkg: update Helm to `v3.1.0-scheme-patched`

hiddeco:
This includes an update of Kubernetes dependencies to `v0.17.2`,
which equals to Kubernetes 1.17.2.


stefanprodan (approved):
LGTM

PR #305, opened 2020-02-14
Update for Helm 3 on OSX using Library

kav:
note the sed can be removed with Helm 3 generally since cache is no longer included


docs/references/helmrelease-custom-resource.md
```diff
@@ -413,6 +413,10 @@ the filename.
 cp ~/.helm/repository/repositories.yaml .
 sed -i -e 's/^\( *cache: \).*\/\(.*\.yaml\)/\1\2/g' repositories.yaml
 ```
+If you are using OSX and Helm 3 the command will be
```
hiddeco:
```suggestion

If you are using OSX and Helm 3 the command will be:

```

to maintain the document style.

hiddeco (approved):

Issue #301, opened 2020-02-13
helm/v3: some error messages are missing formatting and context


derrickburns:
I am seeing error messages (from rc9) like this one that do not have the new release name and namespace fields.
Can we add them?

```
2020/02/13 21:21:00 warning: cannot overwrite table with non table for policy (map[])
```

hiddeco:
Need a bit more context about the operation being performed before I can say if we are able to add them.

derrickburns:
@hiddeco I do not know what is writing the error message into the helm operator logs.  That is the point. Sorry.

derrickburns:
Here is broader context:
```
│ ts=2020-02-13T21:36:01.763355281Z caller=release.go:360 component=release release=loadtest-mongo targetNamespace=loadtest resource=loadtest:helmrelease/mongo helmVersion=v │
│ ts=2020-02-13T21:36:01.76621114Z caller=helm.go:69 component=helm version=v3 info="preparing upgrade for loadtest-mongo" targetNamespace=loadtest release=loadtest-mongo    │
│ ts=2020-02-13T21:36:01.87882374Z caller=helm.go:69 component=helm version=v3 info="performing update for helm-operator" targetNamespace=flux release=helm-operator          │
│ ts=2020-02-13T21:36:01.892692276Z caller=helm.go:69 component=helm version=v3 info="dry run for helm-operator" targetNamespace=flux release=helm-operator                   │
│ ts=2020-02-13T21:36:01.952093947Z caller=release.go:404 component=release release=helm-operator targetNamespace=flux resource=flux:helmrelease/helm-operator helmVersion=v3 │
│ ts=2020-02-13T21:36:02.097241443Z caller=helm.go:69 component=helm version=v3 info="performing update for loadtest-mongo" targetNamespace=loadtest release=loadtest-mongo   │
│ ts=2020-02-13T21:36:02.111553046Z caller=helm.go:69 component=helm version=v3 info="dry run for loadtest-mongo" targetNamespace=loadtest release=loadtest-mongo             │
│ ts=2020-02-13T21:36:02.122426778Z caller=release.go:404 component=release release=loadtest-mongo targetNamespace=loadtest resource=loadtest:helmrelease/mongo helmVersion=v │
│ ts=2020-02-13T21:36:02.167962402Z caller=release.go:360 component=release release=pomerium targetNamespace=pomerium resource=pomerium:helmrelease/pomerium helmVersion=v3 i │
[...]

PR #306, opened 2020-02-16
release: configure go logger to log using our go-kit logger

stefansedich:
Ensure all logging is wrapped using our logger so that the things like the users preferred logging format is respected for all logs even those logging directly to `log.Printf`.

This somewhat tackles #301 but due to the reasons mentioned over there adding log context to these log lines is not possibly without introducing changes over in Helm itself.

Tested this using an unknown hook as seen to be logging using `log.Printf` here https://github.com/helm/helm/blob/840bbc951ea026297a5afeffdf913bd6bc79f20f/pkg/releaseutil/manifest_sorter.go#L192

```
ts=2020-02-16T18:52:09.869319509Z caller=logwriter.go:28 info="2020/02/16 18:52:09 info: skipping unknown hook: \"foo\""
```

### DIFF
--- a/cmd/helm-operator/main.go
+++ b/cmd/helm-operator/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	golog "log"
 	"os"
 	"os/signal"
 	"strings"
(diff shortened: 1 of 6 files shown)


cmd/helm-operator/main.go
```diff
@@ -159,6 +160,10 @@ func main() {
 		logger = log.With(logger, "caller", log.DefaultCaller)
 	}
 
+	// configure go logger to output using go-kit log
+	logWriter := helm.NewLogWriter(logger)
```
hiddeco: Given this is now utilized as a global thing, I think it should be moved to an `utils` package.

stefansedich: @hiddeco done

hiddeco (approved):

Issue #17, opened 2019-04-03
There's not a simple way to add a helm repo with a custom CA


ipedrazas:
It's not easy to add your own repository configuration to the chart.

The setting `Values.helmOperator.configureRepositories.enable` allow you toinject a secret, however, if there's a repo that needs a CA there's not a standard way to add the CA into the deployment provided by the chart.

Note that `Values.helmOperator.tls.verify ` is used as a flag to configure `tiller` tls and not repo.

If you jump into the helmOperator pod and create the file `ca.crt` and then execute:

`helm repo add myrepo https://myhelmrepo.com --ca-file ca.crt`

the `HelmRelease` can use a helm repo with a self-signed cert. This means that the operator needs 2 things to install releases from a helm repo under a self-signed cert: 

* The CA certificate.
* The entry in the `repositories.yaml` with the location of the ca certificate.

hiddeco:
I think this got resolved in #183 (cert path configuration is now possible, and the chart allows you to add additional mounts), and will be further improved with #142.

kav:
Just wanted to leave a note in case anyone else goes down the rabbit hole here. It appears you should do something like the following.


First create a `repositories.yaml` as described in the docs. Add a line for that reads: 
```yaml
  ... 
  caFile: /root/.helm/certs/ca.crt
  ...
```
for any repository using your new ca.

Then run:
`kubectl create secret generic helm-repositories -n flux --from-file=repositories.yaml --from-file=ca.crt`

Then use these values with the helm-operator helm chart (presented as yaml but you can get crazy with command line opts if you want...
```yaml
...
configureRepositories:
  enable: true
  secretName: helm-repositories
extraVolumeMounts:
- name: repositories-yaml
  mountPath: /root/.helm/certs/ca.crt
  subPath: ca.crt
```
obviously you can tweak this for multiple certificate authorities or just move stuff around as you like.

Note: Your HelmRelease should continue to use the url of the repo _not the name_.

PR #311, opened 2020-02-19
Adding configuredRepositories support for Helm 3

### LINKED ISSUE #310: configureRepositories needs the same version check as other locations
**Describe the bug**
configureRepositories needs the same helm3 treatment as the rest of the helm config and cache since they randomly moved the repository files as well.

**To Reproduce**
Steps to reproduce the behavior (using helm.versions=v3)

1. Setup a repositories.yaml file and private key as in: https://github.com/fluxcd/helm-operator/issues/17#issuecomment-586456020 (Also tweaking the locations to match helm3)
1. Notice it doesn't work:
` chart unavailable: looks like "<chart_repo>" is not a valid chart repository or cannot be reached: Get <chart_repo>/index.yaml: x509: certificate signed by unknown authority`
1. Shell into the container and note the repositories.yaml file isn't mounted at `/root/.config/helm/repositories.yaml` (the v3 location) but rather `/root/.helm/repository/repositories.yaml`

kav:
Fixes #310
Adds support for configuredRepositories property when using helm v3

### DIFF
--- a/chart/helm-operator/templates/deployment.yaml
+++ b/chart/helm-operator/templates/deployment.yaml
@@ -155,10 +155,18 @@ spec:
         {{- end }}
         {{- end }}
         {{- if .Values.configureRepositories.enable }}
+        {{- if contains "v2" .Values.helm.versions }}
         - name: {{ .Values.configureRepositories.volumeName | quote }}
           mountPath: /root/.helm/repository/repositories.yaml
           subPath: repositories.yaml
           readOnly: true
+        {{- end -}}
+        {{- if contains "v3" .Values.helm.versions -}}
+        - name: {{ .Values.configureRepositories.volumeName | quote }}
+          mountPath: /root/.config/helm/repositories.yaml
+          subPath: repositories.yaml
+          readOnly: true
+        {{- end -}}
         {{- if contains "v2" .Values.helm.versions }}
         - name: {{ .Values.configureRepositories.cacheVolumeName | quote }}
           mountPath: /var/fluxd/helm/repository/cache


hiddeco (commented):
1. all end-to-end tests are failing due to templating issues
2. I fail to observe why this change is required, given #173 already achieves this and is still in place

kav:
1. Apologies, didn't run the tests, won't happen again. I am seeing a lot of "no test files" on the unit tests however.
2. As far as I can tell #173 doesn't work, at least with a custom ca, didn't test it without.

And further on 2:
From the tone I'm reading some frustration here. Definitely mia culpa for not digging into your testing workflow and committing a typo. Sorry about that. 

Also I apologize if filing issues and PRing them is not the preferred practice. I can wait for verification, start a thread on slack, or just piss off if you'd prefer. I did verify on a running instance that copying the files to the "correct" location resolved the issue I was seeing and it seemed like a straightforward solution.

hiddeco:
My apologies for being brief in communication, due to the requested review I wanted to react quickly but did not take proper time to add some humanity to my message. Hence the frustrated (but not _actually_ frustrated) tone of voice, sorry about that (again). 

**Contributions are absolutely always welcome and much appreciated**, it may however take some time to before someone actually has a look at them, but we do actively monitor submitted issues and PRs -- even without specifically asking for a review. :sunflower:

---

Back to the actual issue: the reason it does not work for you is because you are combining two things that aren't supposed to be combined. The `configureRepositories` is intended to work by adding entries to `configureRepositories.repositories`, [as described here](https://github.com/fluxcd/helm-operator/tree/master/chart/helm-operator#use-a-custom-helm-repository).

I expect things to work without this PR if you:

1. Add the entries to `configureRepositories.repositories` (including the CA path), instead of mounting a custom `repositories.yaml`
2. Mount the CA file using the extra volume mount value

kav:
Here's what I was doing  (https://github.com/fluxcd/helm-operator/issues/17#issuecomment-586456020)
```
...
configureRepositories:
  enable: true
  secretName: helm-repositories
extraVolumeMounts:
- name: repositories-yaml
  mountPath: /root/.helm/certs/ca.crt
  subPath: ca.crt
...
```
I'm not side loading my custom repositories.yaml or anything I'm using the provided secretName option. I suspect if I do the below it still won't work with Helm 3
```
...
configureRepositories:
  enable: true
  secretName: helm-repositories
...
```
Happy to workaround it as suggested if this is all about to be axed in favor of the new crds and maybe I'm misreading the chart... but this really seems like a valid normal use of the configureRepositories.secretName option

hiddeco:
> I'm not side loading my custom repositories.yaml or anything I'm using the provided secretName option. 

I missed this minor detail. Looking at the chart logic, when you do this, the file should be mounted [at the expected path](https://github.com/fluxcd/helm-operator/blob/master/chart/helm-operator/templates/deployment.yaml#L159), and this path is [then used as an argument for the import flag](https://github.com/fluxcd/helm-operator/blob/master/chart/helm-operator/templates/deployment.yaml#L199). I have personally confirmed this to be working when the PR linked earlier was created and am quite sure this should work.

In your issue you mention

> Notice it doesn't work

Can you elaborate a bit more on what signals that it does not work, and include the (boot) logs from the operator and a (redacted) YAML copy of the deployment?

kav:
Will do, I was a bit short there as I didn't notice the import flag and of course saw failures after adding the helm binary and testing so it appeared to just not be plumbed for v3. It seems perhaps I jumped the gun a bit

hiddeco:
@kav how is this going for you now?

kav:
Think I've got it! It works now, so I've been trying to figure out what the differences are. I believe; (haven't had a fully isolated repro yet) that if the chart repo is unavailable or unreachable while the helm operator is coming up it "never recovers", so it complains about the cert, doesn't save the repo in the local file, etc. Killing the pod will recover it.

I was running into that since the helm operator brings up everything in my environment including the chart museum instance it later relies on for further deployments.

kav:
Given all that perhaps we should reject the PR and I'll update the bug with the scoped description and notes from here?

hiddeco:
As you wish! Thank you for putting time an effort in a PR 🍰

PR #322, opened 2020-03-02
install: rename files and resource names

hiddeco:
This PR changes a couple of things.

1. The file names have been changed to just what they represent, as
   the context defining prefixes that used to be there no longer serve
   a real purpose now that the Helm operator has a dedicated
   repository.
2. The names of the deployment and other resources have been simplified,
   the 'Flux branding' adds little to the resources and takes longer to
   type.
3. Some boilerplating for the Tiller certificates has been removed,
   `eksctl` is the sole user of this package and has dropped Helm 2.
4. References to the files have been changed in both `hack/` and
   documentation files.


stefanprodan (approved):
LGTM

PR #324, opened 2020-03-03
allow envFrom secretName

scott-grimes:
flux can easily be configured for [git over HTTPS](https://github.com/fluxcd/flux/tree/master/chart/flux#flux-with-git-over-https). 

This PR allows the same secret (or any secret) to be easily used for both the `flux` and `helm-operator` releases

### DIFF
--- a/chart/helm-operator/README.md
+++ b/chart/helm-operator/README.md
@@ -236,6 +236,7 @@ The following tables lists the configurable parameters of the Flux chart and the
 | `tolerations`                                     | `[]`                                                 | Tolerations properties for the deployment
 | `affinity`                                        | `{}`                                                 | Affinity properties for the deployment
 | `extraEnvs`                                       | `[]`                                                 | Extra environment variables for the Helm Operator pod(s)
+| `env.secretName`                                  | ``                                                   |	Name of the secret that contains environment variables which should be defined in the container (using envFrom)
 | `podAnnotations`                                  | `{}`                                                 | Additional pod annotations
 | `podLabels`                                       | `{}`                                                 | Additional pod labels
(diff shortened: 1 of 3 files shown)


hiddeco:
Thanks for your contribution, any contribution we get is much appreciated :tulip: 

I am however not sure this will work, given the way it works for Flux is through [command substitution that is baked into Kubernetes](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config). As this secret is not available to the `HelmRelease`, it can not be substituted either as far as I know.

scott-grimes:
Yup you're 100% correct, the workaround we ended up using was an initcontainer and shared extraVolume to pull in the https credentials from the flux secret, combined with the git.configEnabled option. 

```
extraVolumes:
  - name: gitcredentials
    emptyDir: {}
extraVolumeMounts:
  - name: gitcredentials
    mountPath: "/root"
initContainers:
  - name: https-git-credhelper
    image: busybox
    envFrom:
      - secretRef:
          name: flux-git-https-auth
    command: ["/bin/sh"]
    args:
      - "-c"
      - 'echo "https://\$(GIT_AUTHUSER):\$(GIT_AUTHKEY)@gitlab.com" > /root/.git-credentials'
    volumeMounts:
      - name: gitcredentials
        mountPath: "/root"
        readOnly: false
git:
  config:
    enabled: true
    data: |
      [credential]
        helper = store
```

hiddeco:
@scott-grimes I think that if you would put that into a `.netrc` file you wouldn't have to configure the credential store in the Git configuration.

PR #326, opened 2020-03-06
status: unset rolledback cond on released=true

hiddeco:
This ensures the condition is always reset when a successful
installation or upgrade happens, and prevents spurious upgrades
from happening due to the rollback condition not being removed
after a rolled back release was manually uninstalled.

Fixes behaviour reported in https://weave-community.slack.com/archives/C4U5ATZ9S/p1583248546169800


stefanprodan (approved):
LGTM

PR #328, opened 2020-03-10
Added hostAliases to deployment template

kammous:
This PR affects Flux Helm Operator chart. It aims to add the ability to statically resolve FQDNs of a Helm charts repository or a git server. By adding a new `hostAliases` parameter to `values.yaml` file, a user can enrich `/etc/hosts` file inside helm operator container by passing his entries to the Helm chart. 
I don't see where this change can be documented. Please let me know if needed.
Thanks


stefanprodan (changes requested):
@kammous please add the field to the chart readme in the configuration section. 

Thanks

kammous:
@stefanprodan Thanks. I've amended the commit to add `hostAliases` field to the chart readme file.

stefanprodan (approved):
LGTM

Thanks @kammous

PR #339, opened 2020-03-24
pkg: Helm 2.16.4

hiddeco:
- [x] Go library `2.16.3` -> `2.16.4`
- [x] Binaries `2.16.3` -> `2.16.4`


stefanprodan:
Helm 2.16.4 is broken as far as I can tell, same error as in https://github.com/fluxcd/flux/issues/2935

hiddeco:
Looks like that indeed, lets wait for `2.16.5`.

stefanprodan:
It's so hard to tell where is the actual error in the e2e tests... can we exit right after an error?

hiddeco:
I often use the information printed after the error to debug things. The errors are repeated `validation failed: ...` errors.

PR #358, opened 2020-04-03
Add missing v1 in apiVersion

hongkunyoo:
Fix little typo


hiddeco (approved):
Excellent catch, thank you :fish:

Issue #370, opened 2020-04-14
Dependency update failure results in upgrade


brew:
**Describe the bug**

I upgraded the Helm-Operator from rc9 to v1 using the command:

```sh
helm repo update
helm upgrade -i helm-operator fluxcd/helm-operator --wait -n fluxcd -f ./hop-values.yaml
```

Where hop-values.yaml contains:

```yaml
git:
  ssh:
    secretName: flux-git-deploy
  pollInterval: 1m

chartsSyncInterval: 1m

configureRepositories:
  enable: true
  repositories:
    - name: bitnami
      url: https://charts.bitnami.com/bitnami
    - name: stable
      url: https://kubernetes-charts.storage.googleapis.com/

helm:
  versions: v3
```

I immediately received cluster errors from the helm-operator for each release along the lines of:

> FailedReleaseSync
> synchronization of release 'myapp-stage' in namespace 'stage' failed: failed to prepare chart for release: chart not ready: git repo has not been cloned yet

Two of my charts have a dependency on the Bitnami mongodb charts. Both of the Mongo pods were stopped, and deployed again shortly after, but their existing PVCs were removed (and data lost).

Helm release status looks like this:
```sh
Status:
  Conditions:
    Last Transition Time:  2020-04-14T08:30:34Z
    Last Update Time:      2020-04-14T08:43:07Z
    Message:               Chart fetch failed for Helm release 'myapp-stage' in 'stage'.
    Reason:                ChartFetchFailed
    Status:                False
    Type:                  ChartFetched
    Last Transition Time:  2020-04-14T08:12:20Z
    Last Update Time:      2020-04-14T08:51:23Z
    Message:               Release was successful for Helm release 'myapp-stage' in 'stage'.
    Reason:                Succeeded
    Status:                True
    Type:                  Released
```

**Expected behavior**

Helm Operator to be updated without affecting existing releases.

**Additional context**

- Helm Operator version: 
```
➜ kubectl images -n fluxcd
[Summary]: 1 namespaces, 3 pods, 3 containers and 3 different images
+---------------------------------+--------------------+--------------------------------------+
|             PodName             |   ContainerName    |            ContainerImage            |
+---------------------------------+--------------------+--------------------------------------+
| flux-c4748d54d-t5hgj            | flux               | docker.io/fluxcd/flux:1.17.1         |
+---------------------------------+--------------------+--------------------------------------+
[...]

PR #372, opened 2020-04-15
release: do not swallow dependency update errors

hiddeco:
Theoretic fix to #370


hiddeco:
Confirmed to be working in https://github.com/fluxcd/helm-operator/issues/370#issuecomment-613968671.

stefanprodan (approved):

PR #387, opened 2020-04-21
Expose the number of Helm Operator managed releases

### LINKED ISSUE #344: Expose a number of Helm Operator managed releases gauge
Hello 👋 

I'd like to get a precise number of Helm Operator managed releases. Using existing summaries is not precise as `count(sum(flux_helm_operator_release_duration_seconds_count) by (namespace, release_name))` also takes uninstalled releases into account since they were observed at some point.

I suggest we expose a new `flux_helm_operator_release_total` (feel free to challenge the name) gauge that tracks the number of releases managed by the Helm Operator.

sa-spag:
Closes #344.

### DIFF
--- a/docs/references/monitoring.md
+++ b/docs/references/monitoring.md
@@ -8,6 +8,7 @@ in Prometheus format.
 
 | Metric | Description
 |--------|---
+| `release_count` | Count of releases managed by the operator. |
 | `release_duration_seconds` | Release synchronization duration in seconds. This duration includes one or many `release_phase_durations`. |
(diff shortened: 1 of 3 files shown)


sa-spag:
Actually I'm wondering if it would make more sense to set the gauge on `InstallAction` and `UninstallAction` from https://github.com/fluxcd/helm-operator/blob/master/pkg/release/release.go#L240, so we track actual releases instead of HelmReleases. Though it won't be enough to take into account previously installed releases.

hiddeco (approved):
:100:

PR #389, opened 2020-04-22
release: don't retry initial install indefinitely upon failure

### LINKED ISSUE #388: Configurability of install retries
**Describe the feature**

Currently upgrades support configurability of retries through the HelmRelease's `rollback.retry` and `rollback.maxRetries` fields. (Side question, is `rollback.retry` necessary, since `rollback.maxRetries` could just default to 0?) However, installs are just infinitely retried, with no configurability. This can make debugging a failed install harder, and can become unnecessary churn in e.g. metrics/alerts.

To enable configurability for install retries, there could be a root-level `maxRetries` field which applies to installs and provides a default value for `rollback.maxRetries` (since this really just specifies how many times to retry an upgrade (with rollbacks instead of uninstalls in between). This field could default to:

* -1, to match the current install behavior (infinite retries)
* 5, to match the the existing `rollback.maxRetries` default
* 6, to match the kubernetes Job backoffLimit default

There could be an ability to override the default at the operator level as well as needed. One could specify e.g. 0 if they wanted to fail as fast as possible, or just reduce down from the default to e.g. 1.

**Statusing**

There could be a `status.installCount` or a `status.retryCount` which also deprecates `status.rollbackCount`, and gets updated when retrying installs or upgrades (when rollbacks are enabled), since those can't happen at the same time (there's just a different strategy to reset the release between retries, either uninstalls or rollbacks). Similar to `status.rollbackCount`, this count would reset when a new change is made to the HelmRelease that needs to be reconciled.

There should also be a new `Uninstalling` value for `status.phase`, which occurs during any release uninstall triggered by the operator:

1. a HelmRelease CR is deleted (only affects the prometheus metrics, since the CR would no longer exist on which to expose `status.phase`)
1. post-install failure when there are more install retries to consume

stefansedich:
Currently when installing a release for the first time upon failure we will retry indefinitely, this change introduces a maximum number of install attempts which now defaults to 5 and introduces a new CRD field if the user wishes to customize the number of retries.

The changes also introduce a new Uninstalling phase and metric which allows users to monitor and alert on it.

Closes #388


pkg/status/metrics.go
```diff
@@ -13,6 +13,7 @@ const (
 
 var phaseToGaugeValue = map[v1.HelmReleasePhase]float64{
 	// Unknown is mapped to 0
+	v1.HelmReleasePhaseUninstallFailed:  -5,
```
hiddeco: Given there has not been a release yet with this new metric, I would reorder the integers so that they make sense again (`ChartFetchFailed` being the lowest negative integer).

stefansedich: Gotcha makes sense.

stefansedich:
@hiddeco @seaneagan ran into an interesting issue with the solution thus far, once failed and uninstalled a release will be in the "Uninstalled" phase in metrics where you can no longer look for it as a failed release.

I could take the approach to flick the metric to failed upon the final uninstall, but how useful will this phase metric be? my testing had it going from uninstalled back to failed in such a short time I would never have the "uninstalled" phase scrapped to make it useful anyway.

Second small issue is for an install that does not actually get to the point of installing anything i.e. a bad chart value, we land in the UninstallFailed phase due to there being no release to uninstall, this one would be an easy fix just to noop if we know there is no release and call it successful anyway.

pkg/status/status.go
```diff
@@ -199,3 +199,12 @@ func ShouldRetryUpgrade(hr *v1.HelmRelease) bool {
...
 	return hr.Spec.Rollback.GetMaxRetries() == 0 || hr.Status.RollbackCount <= hr.Spec.Rollback.GetMaxRetries()
 }
+
+// ShouldRetryInstall returns if the install of a failed release should
+// be retried.
+func ShouldRetryInstall(hr *v1.HelmRelease) bool {
+	if !HasSynced(hr) {
+		return true
+	}
+	return hr.GetMaxRetries() == 0 || hr.Status.InstallCount < hr.GetMaxRetries()
```
seaneagan:
```suggestion
	return hr.GetMaxRetries() < 0 || hr.Status.InstallCount < hr.GetMaxRetries()
```

It would be useful to be able to do 0 retries. To accomplish this, a solution could be to deprecate `rollback.retry` and change default to true, change `rollback.maxRetries: 0` to mean 0 retries, and `rollback.maxRetries: -1` to mean infinite. Same behavior goes for the root-level `maxRetries`, which serves as the default value for `rollback.maxRetries`.

stefansedich: I was avoiding any changes that involved changing anything around rollback, but I see no issue to use 0 and -1 for the root level max retries to turn them off or be infinite but agree it could be confusing.

To me the non "breaking" change would be to add a `retry` field to the root which defaults to `true`, not sure we want to change the existing behavior of rollback retries in a breaking way.

What do you think here @hiddeco ?

seaneagan:
> 
> 
> @hiddeco @seaneagan ran into an interesting issue with the solution thus far, once failed and uninstalled a release will be in the "Uninstalled" phase in metrics where you can no longer look for it as a failed release.

IMO after the final install attempt it should be left in "Failed" status, without doing any uninstall, so that the failed install can be debugged. Potentially that could mean there is no need for an "Uninstalled" phase, if we can just go straight from "Uninstalling" to "Installing" when doing a retry.

> Second small issue is for an install that does not actually get to the point of installing anything i.e. a bad chart value, we land in the UninstallFailed phase due to there being no release to uninstall, this one would be an easy fix just to noop if we know there is no release and call it successful anyway.

Maybe just go straight back to "Installing" (if there are more retries), and skip the "Uninstalling" phase altogether?

One thing I had also considered was whether to include any backoff semantics ala kubernetes Jobs/Pods. Could even go as far as having `backoffLimit` and possibly `activeDeadlineSeconds` fields instead of e.g. `maxRetries`, but I left that out since it would be a shift from the existing rollback API.

stefansedich:
> > @hiddeco @seaneagan ran into an interesting issue with the solution thus far, once failed and uninstalled a release will be in the "Uninstalled" phase in metrics where you can no longer look for it as a failed release.
> 
> IMO after the final install attempt it should be left in "Failed" status, without doing any uninstall, so that the failed install can be debugged. Potentially that could mean there is no need for an "Uninstalled" phase, if we can just go straight from "Uninstalling" to "Installing" when doing a retry.
> 

Ah interesting that can work quite easily, it kind of makes the UninstallXX phases moot as they will live for a short period of time, however they have some use as you can see them in the "Conditions" seeing them being performed so it might have some benefit?

> > Second small issue is for an install that does not actually get to the point of installing anything i.e. a bad chart value, we land in the UninstallFailed phase due to there being no release to uninstall, this one would be an easy fix just to noop if we know there is no release and call it successful anyway.
> 
> Maybe just go straight back to "Installing" (if there are more retries), and skip the "Uninstalling" phase altogether?
> 

Yup if there is no current release I can skip the uninstalling altogether as there is nothing to uninstall, makese perfect sense. I am going to implement this and the above fix as they make sense and solve those two issues.

> One thing I had also considered was whether to include any backoff semantics ala kubernetes Jobs/Pods. Could even go as far as having `backoffLimit` and possibly `activeDeadlineSeconds` fields instead of e.g. `maxRetries`, but I left that out since it would be a shift from the existing rollback API.

I agree here that not changing the existing API too much makes the most sense.

pkg/status/status.go
seaneagan: After a little more thought, an `uninstall` field could be more consistent with the existing `rollback` field, with the exception of a couple defaults being different:

```yaml
spec:
  uninstall:
    enable: true # default true
    retry: true # default true
    maxRetries: 1 # default 5
    disableHooks: true # default false
    timeout: 300 # default 300
  rollback: # no changes
    enable: true # default false
    retry: true # default false
    maxRetries: 1 # default 5
    disableHooks: true # default false
    timeout: 300 # default 300
```

and then instead of `status.installCount` it could be `status.uninstallCount` to match `status.rollbackCount`.

Only thing is I'm not sure if there could be some confusion about whether `uninstall` references configuration for deleting the CRD. If this is a problem there could be another field called `uninstall.enableOnDelete` or similar (default true) to account for this. Or change `uninstall.enable` to `uninstall.enableOnInstallFailed`.

hiddeco: I like the above proposal, and the spec fits in to the picture I had in mind for a future gen API with finer grain option control.

stefansedich: Do the semantics of "uninstall" make sense to a user however? I still think of it as retrying the install and not the uninstalls, in my mind I care about retrying the install if it fails, not the uninstall.

And when say `uninstall.enable=false` do we now not uninstall when removing a HR for example?

hiddeco: > And when say `uninstall.enable=false` do we now not uninstall when removing a HR for example?

We do now uninstall when an HR is removed, and I would expect this behaviour to stay put. Due to this, I agree "install" is a better option as the flag and logic applies to this process.

stefansedich: We said we cannot use "install" right? I know when we talked about it you had said we cannot as it is seen as the spec root. I had not tried it yet but just recall that.

stefansedich: Okies after chatting with @hiddeco offline the plan is to add the retry related fields into an "install" section.

This change will also pave the way to move other "install" related fields into a logical grouping in the future moving them out of the spec root and grouping things logically.

Thoughts on that plan @seaneagan ?

```
spec:
  install:
    retry: true # default true
    maxRetries: 1 # default 5
  rollback: # no changes
    enable: true # default false
    retry: true # default false
```

seaneagan: Yes, I agree that it is really about the desire to retry installs, not to do uninstalls, and install retries should imply uninstalls. But similarly some of the rollback fields are about the desire to retry upgrades, so I think similarly it would make sense to deprecate those fields and replace them with an `upgrade` key, which I think could be done in a backward compatible way. And if moving them, should consider removing the `retry` field, and just allowing `maxRetries=0` to mean no retries. Also it may be nice to be able to define whether to do a rollback/uninstall after all retries are consumed (or leave release in failed state for debugging).

```yaml
spec:
  install:
    maxRetries: 1 # default 0 (none), >0 implies uninstalls
    uninstallAfterMaxRetries: true # default false
    noHooks: true
  upgrade:
    maxRetries: 1 # default 0 (none), >0 implies rollbacks
    rollbackAfterMaxRetries: true # default false
    force: true
    noHooks: true
  rollback: # how to do rollbacks, not whether to
    force: true
    noHooks: true
    timeout: 300
  uninstall: # how to do uninstalls, not whether to 
    noHooks: true
    timeout: 300
```

`rollback.enable/retry/maxRetries` would all be deprecated, and it would be an error to specify any of them at the same time as specifying an `upgrade` key. It may make sense to also deprecate `status.rollbackCount` for `status.upgradeRetryCount` and add `status.installRetryCount`.

This is somewhat separate, but currently `wait` becomes true (even if it is explicitly specified as false) for upgrades when rollbacks are enabled. I think it should at least use the value of `wait` when explicitly specified, and probably rollbacks shouldn't change the default value of `wait`.

stefansedich: @seaneagan that all seems sensible, but definately something I won't tackle as part of this PR, it has already grown a bit hairier than I anticipated in some of the edge cases but I can definately see a future PR after this to go through everything and improve the semantics behind all the existing fields.

pkg/apis/helm.fluxcd.io/v1/types_helmrelease.go
```diff
@@ -256,6 +256,35 @@ type ExternalSourceSelector struct {
 	Optional *bool `json:"optional,omitempty"`
 }
 
+type Install struct {
+	// Retry will mark this Helm release for upgrade retries after a
```
hiddeco: Comment does not match the flag behaviour (`/s/upgrade/install/`, `/s/rollback/installation failure/`)

pkg/release/release.go
```diff
@@ -379,10 +402,12 @@ func (r *Release) install(client helm.Client, hr *v1.HelmRelease, chart chart, v
 		Wait:       hr.Spec.Wait,
 	})
 	if err != nil {
+		status.SetInstallPhase(r.hrClient.HelmReleases(hr.Namespace), hr, v1.HelmReleasePhaseFailed)
 		status.SetStatusPhase(r.hrClient.HelmReleases(hr.Namespace), hr, v1.HelmReleasePhaseFailed)
```
hiddeco: Obsolete / duplicate of the above?

pkg/release/release.go
```diff
@@ -379,10 +402,12 @@ func (r *Release) install(client helm.Client, hr *v1.HelmRelease, chart chart, v
...
 		err = fmt.Errorf("installation failed: %w", err)
 		return
 	}
+	status.SetInstallPhase(r.hrClient.HelmReleases(hr.Namespace), hr, v1.HelmReleasePhaseSucceeded)
```
hiddeco: This is technically also a duplicate of the call below, except that it does not push the revision. I think this can be prevented by resetting the counter on a `v1.HelmReleasePhaseSucceeded` as is already done for successful upgrades after a rollback.

pkg/apis/helm.fluxcd.io/v1/types_helmrelease.go
```diff
@@ -545,6 +591,12 @@ type HelmReleaseStatus struct {
 	// +optional
 	LastAttemptedRevision string `json:"lastAttemptedRevision,omitempty"`
 
+	// FailedInstallCount records the number of failed installs,
+	// it is incremented after a install failure and reset after a
```
hiddeco:
```suggestion
	// it is incremented after an install failure and reset after a
```

pkg/release/release.go
```diff
@@ -270,14 +270,30 @@ next:
...
+
 		logger.Log("info", "running installation", "phase", action)
 		newRel, err = r.install(client, hr, chart, values)
 		if err != nil {
 			logger.Log("error", err, "phase", action)
 			errs = append(errs, err)
 
-			action = UninstallAction
-			goto next
+			if status.ShouldRetryInstall(hr, hr.Status.FailedInstallCount+1) {
```
hiddeco: Just checking on logic: you are passing the count here as it was increased by the unsuccessful installation but the HR does not reflect this?

pkg/release/release.go
stefansedich: This one is right, there are 3 phases:

Installing: Clear count to 0
Failed: Increment
Success: Clear count to 0

I could just re-use the `HelmReleasePhaseSucceeded` however in the clearing case if desired.

pkg/release/release.go
stefansedich: I am not sure I follow, failed will increment and success/install reset?  I am not seeing any duplicate calls?

pkg/release/release.go
stefansedich: Correct, the way the phase setting works means it will not "actually" be incremented until the next sync loop.

pkg/release/release.go
hiddeco: `SetInstallPhase` and `SetStatusPhaseWithRevision` do the same thing: they update the status of the HelmRelease with the given phase, except the install version takes the reset and increment of the counter into account.

I think this reset logic can be moved to `SetCondition` so that `SetInstallPhase` no longer needs to be called here as no additional context is required for the reset (but only for the increment on failure).

seaneagan:
@stefansedich is this ready for review (still marked as draft)?

hiddeco:
@seaneagan this is waiting for some test coverage but Stefan has some personal things to take care of first before OSS has prio again. :rose:

stefansedich:
> @stefansedich is this ready for review (still marked as draft)?

Yup sorry for dropping the ball on this one @seaneagan, have been out of action here for a bit. But I am ready to pick things back up again, will get back to adding some e2e tests this week/end and get this one ready for review.

seaneagan:
> Yup sorry for dropping the ball on this one @seaneagan, have been out of action here for a bit. But I am ready to pick things back up again, will get back to adding some e2e tests this week/end and get this one ready for review.

No rush, I just wanted to ensure you weren't waiting for code review, since it's already looking really good. Thanks for working on this!

kingdonb:
Closing, stale.

Thanks for contributing here! Flux v1 and Helm Operator are in maintenance mode 🎉

PR #400, opened 2020-05-02
Fix e2e tests on MacOS.

vladlosev:
This change fixes e2e tests on MacOS. Some of the tests use the `sed` command with the `-i` option to modify files in place. On Linux, that option has an optional parameter specifying the name of the back-up file to create. On the standard `sed` installation in MacOS, the parameter for `-i` is required (it may be empty though, in which case no back-up file is created), so the tests fail with this message:
```
# (in test file ./15_upgrade.bats, line 67)
#   `sed -i 's%replicaCount: 1%replicaCount: 2%' charts/podinfo/values.yaml' failed
```
 This change modifies sed invocation to work on MacOS. This should help with the development of Helm Operator on Mac.

### DIFF
--- a/test/e2e/15_upgrade.bats
+++ b/test/e2e/15_upgrade.bats
@@ -1,5 +1,13 @@
 #!/usr/bin/env bats
 
+function sed_in_place() {
+  if test "$(uname)" = Darwin ; then
+    sed -i '' "$@"
+  else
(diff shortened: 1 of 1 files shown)


hiddeco (approved):

hiddeco:
Thank you @vladlosev :100:

PR #401, opened 2020-05-03
Add `priorityClassName` value to Helm Operator

rowanj:
This is necessary in overcontended environments to help ensure the operator does not get evicted, or to allow it to preempt other workloads.

--

Pulled basically verbatim from the fluxcd chart.


rowanj:
The circleci failure seems to be a timeout... ? Not really familiar with that yet but I'll take a peek...

stefanprodan (approved):
LGTM

PR #402, opened 2020-05-04
[Chart] Stopping ClusterRole/ClusterRoleBinding keep created when it shouldn't

swade1987:
Fixes the issue whereby the ClusterRole and ClusterRoleBinding was being created when the following was specified in the chart values.

```
clusterRole:
  create: false
```


stefanprodan (approved):
LGTM

squaremo:
There seem to be two ClusterRoles at stake:

 - one called `<whatever>-crd` that's created if you `--set clusterRole.create: false`
 - one called `<whatever>` that's created if you `--set clusterRole.create: true`

The first specifically has permissions for apiextensions (CRDs) only. This makes me think having that is essential to the running of the helm-operator, so removing it altogether might cause problems.

stefanprodan:
@squaremo the `<whatever>-crd` cluster role got in this chart by mistake, I forgot to delete it when I've split the Flux chart. Helm-op doesn't use the discovery API nor does it reads CRDs/namespaces.

squaremo:
I see, OK. The code [makes reference to Helm3 using the apiextensions API](https://github.com/fluxcd/helm-operator/blob/master/pkg/helm/v3/helm.go#L48) but I did not chase down how and when.

The other consideration here is that this PR changes behaviour: before, setting 

    --set clusterRole.create=false --set clusterRole.name=foobar

would create a ClusterRoleBinding to `foobar`. Now, the binding will be omitted, given those settings.

The relevant stanza in values.yaml is this:

```
# If create is `false` and no name is given, the Helm
# Operator will be restricted to the namespace where it is
# deployed, and the kubeconfig default context will be set to that
# namespace.
clusterRole:
  create: true
  # The name of a cluster role to bind to; if not set and create is
  # true, a name based on fullname is generated
  name:
```

.. which does not make it clear what the intended meaning of `--set clusterRole.create --set clusterRole.name=foobar` is -- but in any case, this PR changes the outcome.

swade1987:
@squaremo I completely agree with you, however I feel the current values file is somewhat misleading as well because you set `clusterRole.create: false` and you still get a ClusterRole.

My ideal proposal with this is to make sure if you set `clusterRole.create: false` then you get a Role and RoleBinding scoped to the namespace you specified.

squaremo:
> My ideal proposal with this is to make sure if you set `clusterRole.create: false` then you get a Role and RoleBinding scoped to the namespace you specified.

As it stands, if _all_ you do is `--set clusterRole.create=false`, that's what will happen (modulo the spurious *-crd clusterrole).

It looks like `clusterRole.name` was ported from the flux chart, with motivation explained here:
https://github.com/fluxcd/flux/commit/35fac108e8a84cd5f83d1ddeebbd900f2a158325

> If you are stamping the Helm chart into a cluster a few times, it's
> good to be able to refer to a single cluster role, rather than
> creating one for each stamping.

I don't think this situation applies to the Helm operator, since you _either_ run many with namespaced roles, _or_ one with a clusterrole. If so, the name setting doesn't have a use, and should be removed. But it would be less likely to break things to just leave the name setting and honour it when `create: true`. So I'd accept just changing the comment in values.yaml so it's more precise.

swade1987:
So what needs to be changed, I will try it my side and make sure the resources created are what we are expecting.

chart/helm-operator/templates/rbac.yaml
```diff
@@ -20,8 +19,6 @@ rules:
       - '*'
     verbs:
       - '*'
-{{- end -}}
-{{- if or .Values.clusterRole.create .Values.clusterRole.name }}
```
squaremo, line 24: The different conditional here means that using

    --set clusterRole.create=false --set clusterRole.name=foobar

will have a different result to before. I think this is OK as a change, as mentioned in the comments section; please update the description of these settings in `values.yaml` so it's clear that `clusterRole.create: false` will now suppress the ClusterRole _and_ ClusterRoleBinding.

squaremo (changes requested):

chart/helm-operator/templates/rbac.yaml
swade1987: gotcha, will make that change in the next few minutes.

swade1987: Done.

chart/helm-operator/values.yaml
```diff
@@ -144,10 +144,9 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name:
 
-# If create is `false` and no name is given, the Helm
-# Operator will be restricted to the namespace where it is
-# deployed, and the kubeconfig default context will be set to that
-# namespace.
+# If create is `false`  the Helm Operator will be restricted to the namespace
```
squaremo, line 147: :100:

squaremo (approved):

squaremo:
@stefanprodan Can you check you're happy with the change to the meaning of `clusterRole.create` please?

stefanprodan:
Looks ok 👍

squaremo:
Great -- Steve, if you rebase this on master we can merge it. Thank you!

PR #414, opened 2020-05-11
add SecurityContext for Pod

alemairebe:
When working with restrictive PSP, it enables to enforce lower permission with the securitycontext. A small example is provided in values.yaml


hiddeco:
This change has the same conflicts as the ones described in https://github.com/fluxcd/flux/pull/3051.

kingdonb:
Closing, stale.

Thanks for contributing here! Flux v1 and Helm Operator are in maintenance mode 🎉

PR #421, opened 2020-05-19
Allow selecting cross-namespace secrets for Git chart source credentials

### LINKED ISSUE #420: Allow selecting cross-namespace secrets for Git chart source credentials
**Describe the feature**
As mentioned in this comment https://github.com/fluxcd/helm-operator/pull/172#issuecomment-576766725, it would be desirable to be able to fetch the credentials for a Git chart source from secrets in a namespace other that the one of the HelmRelease. 

An example use case can be when having multiple releases in different namespaces, but using the same shared git credentials to read the chart source. In the current version these credentials would need to be duplicated with the same content in all these namespaces.

Example of the desired configuration could be just adding `namespace` to the `secretRef`:|

```yaml
. . .
spec:
  chart:
    git: https://github.com/org/repo
    ref: master
    path: charts/podinfo
    secretRef:
      name: git-https-credentials
      namespace: common # <==== new field
. . .
```

fllaca:
Closes #420. It just adds a `namespace` field to the `secretRef` block of the Git chart source configuration:

```yaml
. . .
spec:
  chart:
    git: https://github.com/org/repo
    ref: master
    path: charts/podinfo
    secretRef:
      name: git-https-credentials
      namespace: common # <==== new field
. . .
```

 This is my very first PR to this project, please be gentle! :P Let me know if there's anything missing that should be added to the PR.


pkg/apis/helm.fluxcd.io/v1/types.go
```diff
@@ -18,6 +18,11 @@ type LocalObjectReference struct {
 	Name string `json:"name"`
 }
 
+type ObjectReference struct {
```
hiddeco, line 21:
```suggestion
type NamespacedObjectReference struct {
```

may be more fitting.

hiddeco (commented):
Looking good, one minor comment about naming :star:

pkg/apis/helm.fluxcd.io/v1/types.go
hiddeco: Never mind, looking at the Kubernetes source `ObjectReference` is fine as well.

hiddeco (approved):

PR #422, opened 2020-05-20
helm/v3: garbage collect anonymous index files

hiddeco:
The Helm 3 dependency update generates anonymous index files on every
run without garbage collecting them, resulting in spurious disk usage.
Configuring a temporary cache directory and garbage collecting it after
the dependency update is not possible as the download manager seems to
ignore the configured repository cache path and always create the index
file in the XDG cache directory.

Instead we detect the any anonymous index filex by their "=" suffix and
delete if they are older than 2m.

### DIFF
--- a/pkg/helm/v3/dependency.go
+++ b/pkg/helm/v3/dependency.go
@@ -1,12 +1,22 @@
 package v3
 
 import (
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
 	"helm.sh/helm/v3/pkg/downloader"
 
 	"github.com/fluxcd/helm-operator/pkg/utils"
 )
 
 func (h *HelmV3) DependencyUpdate(chartPath string) error {
+	// Garbage collect before the dependency update so that
+	// anonymous files from previous runs are cleared, with
+	// a safe guard time offset to not touch any files in
+	// use.
+	garbageCollect(repositoryCache, time.Second * 120)
 	out := utils.NewLogWriter(h.logger)
 	man := &downloader.Manager{
 		Out:              out,
@@ -16,4 +26,22 @@ func (h *HelmV3) DependencyUpdate(chartPath string) error {
 		Getters:          getterProviders(),
 	}
 	return man.Update()
+}
+
+// garbageCollect walks over the files in the given path and deletes
+// any anonymous index file with a mod time older than the given
+// duration.
+func garbageCollect(path string, olderThan time.Duration) {
+	now := time.Now()
+	filepath.Walk(path, func(p string, f os.FileInfo, err error) error {
+		if err != nil || f.IsDir() {
+			return nil
+		}
+		if strings.HasSuffix(f.Name(), "=-index.yaml") || strings.HasSuffix(f.Name(), "=-charts.txt") {
+			if now.Sub(f.ModTime()) > olderThan {
+				return os.Remove(p)
+			}
+		}
+		return nil
+	})
 }


pkg/helm/v3/dependency.go
```diff
@@ -1,12 +1,18 @@
...
+	"strings"
+	"time"
+
 	"helm.sh/helm/v3/pkg/downloader"
 
 	"github.com/fluxcd/helm-operator/pkg/utils"
 )
 
 func (h *HelmV3) DependencyUpdate(chartPath string) error {
+	defer garbageCollect(repositoryCache, time.Second * 120)
```
squaremo: This is cleaning up files from _previous_ runs, so there's no reason to wait until after this run (it's not going to pick up any more files). I would remove the `defer`, and put a comment explaining why this is a convenient place to call it.

pkg/helm/v3/dependency.go
```diff
@@ -17,3 +23,22 @@ func (h *HelmV3) DependencyUpdate(chartPath string) error {
...
 	return man.Update()
 }
+
+// garbageCollect walks over the files in the given path and deletes
+// any anonymous index file with a mod time older than the given
+// duration.
+func garbageCollect(path string, olderThan time.Duration) {
+	now := time.Now()
+	filepath.Walk(path, func(p string, f os.FileInfo, err error) error {
+		if f.IsDir() {
```
squaremo: Do you need to check `err` first?

squaremo (commented):
Comments 💬!

pkg/helm/v3/dependency.go
```diff
@@ -1,12 +1,22 @@
...
+	"strings"
+	"time"
+
 	"helm.sh/helm/v3/pkg/downloader"
 
 	"github.com/fluxcd/helm-operator/pkg/utils"
 )
 
 func (h *HelmV3) DependencyUpdate(chartPath string) error {
+	// Garbage collect before the dependency update so that
```
squaremo, line 15: :100: thank you

pkg/helm/v3/dependency.go
```diff
@@ -17,3 +27,22 @@ func (h *HelmV3) DependencyUpdate(chartPath string) error {
...
+// any anonymous index file with a mod time older than the given
+// duration.
+func garbageCollect(path string, olderThan time.Duration) {
+	now := time.Now()
+	filepath.Walk(path, func(p string, f os.FileInfo, err error) error {
+		if err != nil || f.IsDir() {
+			return nil
+		}
+		if strings.HasSuffix(f.Name(), "=-index.yaml") || strings.HasSuffix(f.Name(), "=-charts.txt") {
+			if diff := now.Sub(f.ModTime()); diff > olderThan {
```
squaremo: Could this just be `if now.Sub(f.ModTime()) > olderThan`? (perhaps the diff value was used, in an earlier version?)

squaremo (approved):
LGTM 🍇

pkg/helm/v3/dependency.go
hiddeco: You are right, some previous version did rely on a `time.Time` I think (and I thus needed to make a transformation to a comparable format iirc).

ferrandinand:
I updated to helm-operator 1.1.0 but this problem is still there as `-index.yaml` and `-charts.txt` are not removed.

```
/ # du -h -s /root/.cache/helm/repository
10.1G   /root/.cache/helm/repository
```
```
  flux-helm-operator:
    Container ID:  docker://b4457b2bf530ab1871441c573d3804ad146946338838f4639287f2a75119d40c
    Image:         docker.io/fluxcd/helm-operator:1.0.1
    Image ID:      docker-pullable://fluxcd/helm-operator@sha256:a2a8bf7538458291ebde17a1f9e5ec0d711f6c9f0a3011e6eb9da680e5b82f61
    Port:          3030/TCP
    Host Port:     0/TCP
    Args:
      --enabled-helm-versions=v3
      --log-format=fmt
      --git-timeout=20s
      --git-poll-interval=5m
      --charts-sync-interval=3m
      --status-update-interval=30s
      --update-chart-deps=true
      --log-release-diffs=false
      --workers=4
      --tiller-namespace=kube-system
```

```
/ # ls -la /root/.cache/helm/repository
total 10576636
-rw-r--r--    1 root     root          3370 Jun  3 16:32 +1EK02SqaEAwVsIXA3tl8lPK64s=-charts.txt
-rw-r--r--    1 root     root       8493397 Jun  3 16:32 +1EK02SqaEAwVsIXA3tl8lPK64s=-index.yaml
-rw-r--r--    1 root     root          3370 Jun  3 17:28 +7TDZRyYcW3jM1iRy-awgDNFAbs=-charts.txt
-rw-r--r--    1 root     root       8493388 Jun  3 17:28 +7TDZRyYcW3jM1iRy-awgDNFAbs=-index.yaml
-rw-r--r--    1 root     root          3370 Jun  4 09:44 +81k-ha-3bQVj-DLN-2sD37Ucmk=-charts.txt
-rw-r--r--    1 root     root       8497898 Jun  4 09:44 +81k-ha-3bQVj-DLN-2sD37Ucmk=-index.yaml
-rw-r--r--    1 root     root          3370 Jun  3 17:26 +BDESiuNv-fjf29DQYP2mb-MhTY=-charts.txt
-rw-r--r--    1 root     root       8493388 Jun  3 17:26 +BDESiuNv-fjf29DQYP2mb-MhTY=-index.yaml
-rw-r--r--    1 root     root          3370 Jun  4 01:20 +DXau1YTZbBln-J7Z-HmErmGCkU=-charts.txt
-rw-r--r--    1 root     root       8495036 Jun  4 01:20 +DXau1YTZbBln-J7Z-HmErmGCkU=-index.yaml
-rw-r--r--    1 root     root          3370 Jun  4 08:01 +HmzFV50LLmf0uyxUxWC6wEcVPo=-charts.txt
[OMITED FULL OUTPUT]
```

hiddeco:
@ferrandinand

```
  flux-helm-operator:
    Container ID:  docker://b4457b2bf530ab1871441c573d3804ad146946338838f4639287f2a75119d40c
    Image:         docker.io/fluxcd/helm-operator:1.0.1
```

Clearly states you are still running `1.0.1`?

ferrandinand:
me 👎.

Thanks @hiddeco

PR #425, opened 2020-05-21
release: excl rolled back rel from dry-run success

hiddeco:
Follow up on #385 to restore rollback behaviour.


stefanprodan (approved):
LGTM

PR #431, opened 2020-05-27
Set `release_condition_info` labels properly

sa-spag:
Hey there 👋

Both `target_namespace` and `release_name` were wrongly set using the HelmRelease metadata. It may be worth releasing this as a patch.


hiddeco (approved):
This will likely make it to a patch release but I can't give you an exact time or date.

PR #433, opened 2020-05-27
Fix comment regarding repositories.yaml setup

MMartyn:
I hit this while enabling gitops via eksctl and recalled it being fixed in the charts at one point:
https://github.com/fluxcd/helm-operator/blob/master/chart/helm-operator/templates/deployment.yaml#L171


hiddeco:
@MMartyn can you please run `make generate` so that all the things depending on the template are rebuild and include the change?

MMartyn:
Updated. Sorry, should have thought about that.

hiddeco (approved):

PR #434, opened 2020-05-28
Migrate chart publishing to GitHub Actions

stefanprodan:
Use the [helm-gh-pages](https://github.com/stefanprodan/helm-gh-pages) GitHub Action for publishing the chart to GitHub Pages.

### DIFF
--- a/.github/workflows/chart-release.yml
+++ b/.github/workflows/chart-release.yml
@@ -1,0 +1,19 @@
+name: release-chart
+on:
+  push:
+    tags: 'chart-*'
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/chart-release.yml
```diff
@@ -0,0 +1,19 @@
+name: release-chart
+on:
+  push:
+    tags: 'chart.*'
```
hiddeco: I think for GitHub Actions this should be `chart-*`.

hiddeco (approved):

PR #451, opened 2020-06-07
Align Helm chart README and values.yaml

chrisjholly:
A number of small changes to the helm chart documentation that I have noticed (and needed) as I was configuring our helm operator over the last week. I hope this is the last PR on helm chart documentation but I feel would be useful to anyone new to quick reference via the readme.


hiddeco (approved):
LGTM, thank you @chrisjholly :cake:

PR #455, opened 2020-06-09
Use sigs.k8s.io/yaml everywhere

hezhizhen:


hiddeco (commented):
Good catch, and likely the result of some past merge-conflict. Looking at the Helm source however, `sigs.k8s.io/yaml` is used. Given this, I think the change should be made in `values.go` and not in the test.

hezhizhen:
> Good catch, and likely the result of some past merge-conflict. Looking at the Helm source however, `sigs.k8s.io/yaml` is used. Given this, I think the change should be made in `values.go` and not in the test.

So I should replace `github.com/ghodss/yaml` with `sigs.k8s.io/yaml` not vice versa?

hiddeco:
Correct.

hezhizhen:
Done.

hiddeco (approved):
Thank you @hezhizhen :tulip:

Issue #452, opened 2020-06-08
Occasional helm test runs after deployment


seaneagan:
As an extension to #369, it may be useful to occasionally run helm tests on a deployed release, both on demand and periodically:

1. Add `helm.fluxcd.io/testAt: <timestamp>` annotation, similar to #453, except only triggers tests, not a full sync.
2. Add `spec.test.interval` field which would cause the helm tests to be re-run if they haven't been run within the interval.

Unclear if these tests should result in uninstalls/rollbacks/retries. Probably not by default, maybe useful though as configuration option.

hiddeco:
If we opt to introduce `helm.fluxcd.io/testAt` (for which you got my vote), we should introduce `helm.fluxcd.io/syncAt` as well to clear the path for #444.

hiddeco:
Ah, I had not seen #453 yet. :ok_hand:

hiddeco:
One thing that is up for debate, given the behaviour introduced in #369, is the rollback. I am not sure that if you run tests periodically to assert everything is (still) working as expected, you want a rollback to happen if it fails several days after the actual release was made.

seaneagan:
:+1: , added #454 for configuration of rollbacks/uninstalls on test failures

gmtstephane:
Hi, what about the initial message ?
Will it be possible to trigger test on a specific release ?
Or to configure a test to run every X minutes ?

Thanks

kingdonb:
Sorry if your issue remains unresolved. The Helm Operator is in maintenance mode, we recommend everybody upgrades to Flux v2 and Helm Controller.

A new release of Helm Operator is out this week, 1.4.4.

We will continue to support Helm Operator in maintenance mode for an indefinite period of time, and eventually archive this repository.

Please be aware that Flux v2 has a vibrant and active developer community who are actively working through minor releases and delivering new features on the way to General Availability for Flux v2.

In the mean time, this repo will still be monitored, but support is basically limited to migration issues only. I will have to close many issues today without reading them all in detail because of time constraints. If your issue is very important, you are welcome to reopen it, but due to staleness of all issues at this point a new report is more likely to be in order. Please open another issue if you have unresolved problems that prevent your migration in the appropriate Flux v2 repo.

[...]

PR #463, opened 2020-06-17
Adding support for optional rollback when helm test fails

### LINKED ISSUE #454: Ability to ignore (or not) helm test failures
As an extension to #369 it would be useful to be able to run helm tests in a non-gating fashion, without an uninstall/rollback on failure, primarily when there is a known issue with a helm test or it is newly implemented and not yet trusted, so that you can still see the test result and logs for debugging purposes without it affecting the deployment.

```
spec:
  test:
    enable: true
    ignore: true # default false
    ignoreDelayed: false # default true
```

`ignoreDelayed` would configure the behavior for tests triggered on a delayed basis as per #452

For ignored tests failures, the `Tested` condition would be `False`, but `Released` would remain `True`.

chrisjholly:
Adding failing e2e test for helm test integration that does not rollback the release on a test failure.

fixed #454


docs/helmrelease-guide/tests.md
```diff
@@ -23,6 +23,18 @@ spec:
...
+## Rollback release on test failure
+
+RollbackOnFailure will cause a Helm release to be rolled back if the tests fails otherwise it 
+will be left in a released state. rollbackOnFailure is optional and defaults to true 
+
+```yaml
+spec:
+  test:
+    enable: true
+    rollbackOnFailure: true
```
seaneagan: There are two scenarios to consider:

1. test fails after install, uninstall?
2. test fails after upgrade, rollback?

I think in most cases one would want the same answer to both of these questions. Also I think anytime a corrective action is taken (uninstall / rollback), the `Released` condition should be marked `false`. So I think a field name which takes all of that into account e.g. `ignore` or `gating` (opposite polarity) may make more sense. Also note the potential for a separate field to account for #452 e.g. `ignoreDelayed`, so that should be considered here.

pkg/release/release.go
```diff
@@ -306,18 +306,31 @@ next:
...
+				err = fmt.Errorf("test failed: %w", err)
+
 				logger.Log("error", err, "action", TestAction)
 				errs = append(errs, err)
 
-				if curRel == nil {
-					action = UninstallAction
+				if hr.Spec.Test.GetRollbackOnFailure() {
+					status.SetStatusPhase(r.hrClient.HelmReleases(hr.Namespace), hr, v1.HelmReleasePhaseFailed)
+					status.SetStatusPhase(r.hrClient.HelmReleases(hr.Namespace), hr, v1.HelmReleasePhaseTestFailed)
```
seaneagan: The `Tested` condition should be marked `false` any time a test fails, so that the user becomes aware, only the overall `Released` condition should be dependent on the rollback / uninstall behavior.

pkg/status/conditions.go
```diff
@@ -129,15 +129,9 @@ func ConditionsForPhase(hr *v1.HelmRelease, phase v1.HelmReleasePhase) ([]v1.Hel
...
 	case v1.HelmReleasePhaseTestFailed:
-		message := fmt.Sprintf(`Test failed for Helm release '%s' in '%s'.`, hr.GetReleaseName(), hr.GetTargetNamespace())
 		condition.Type = v1.HelmReleaseTested
 		condition.Status = v1.ConditionFalse
-		condition.Message = message
-		conditions = append(conditions, &v1.HelmReleaseCondition{
-			Type:    v1.HelmReleaseReleased,
-			Status:  v1.ConditionFalse,
-			Message: message,
-		})
```
seaneagan: Rather than removing this, I'd think we'd want to just wrap it in an `if` here which checks `hr.Spec.Test.RollbackOnFailure` (or whatever field that becomes).

docs/helmrelease-guide/tests.md
chrisjholly: Thanks for your initial feedback @seaneagan  - I have listed below my thoughts:

- At the time, I felt that rollback describe the process of reverting to the last known state (if that is uninstalling as there is no previous state or reverting to a previous release.
- I called it "rollbackOnFailure" as I did not feel 'ignore' gave it enough context. 

Would you be happy with 'ignoreFailures' which would also allow for 'ignoreDelayedFailures'?

p.s this is my first code contribution to helm-operator so I'm new here. It's also the first time I have written Golang so code feedback is welcome. I have another commit adding further code coverage but I am still mastering the go testing framework

seaneagan (commented):
`ignoreFailures` does add a bit of clarity, I'd be curious what others ( @hiddeco ) think of whether `ignore` is clear enough. Since we already have `enable` to turn on/off tests altogether, it seems clear to me that `ignore` would be about ignoring the test results.

pkg/status/conditions.go
chrisjholly: agree - done

docs/helmrelease-guide/tests.md
hiddeco: `ignoreFailures` is perfect :ok_hand:

hiddeco (commented):
Thank you for your work on this @chrisjholly :100: 

The changes look alright to me, but the PR itself could use some commit cleaning and a rebase on `master` before it can be merged.

pkg/status/conditions.go
```diff
@@ -133,11 +133,19 @@ func ConditionsForPhase(hr *v1.HelmRelease, phase v1.HelmReleasePhase) ([]v1.Hel
 		condition.Type = v1.HelmReleaseTested
 		condition.Status = v1.ConditionFalse
 		condition.Message = message
-		conditions = append(conditions, &v1.HelmReleaseCondition{
-			Type:    v1.HelmReleaseReleased,
-			Status:  v1.ConditionFalse,
-			Message: message,
-		})
+		if hr.Spec.Test.GetIgnoreFailures() {
```
seaneagan, line 136: condition needs to be negated

pkg/apis/helm.fluxcd.io/v1/types_helmrelease.go
```diff
@@ -336,6 +340,17 @@ type Test struct {
 	Cleanup *bool `json:"cleanup,omitempty"`
 }
 
+// IgnoreFailures returns the configured ignoreFailures flag,
+// or the default of true to preserve backwards compatible
+func (t Test) GetIgnoreFailures() bool {
+	switch t.IgnoreFailures {
+	case nil:
+		return true
```
seaneagan, line 348: should default to false (not ignoring test failures)

pkg/release/release.go
```diff
@@ -306,22 +364,32 @@ next:
...
+				err = fmt.Errorf("test failed: %w", err)
+
 				logger.Log("error", err, "action", TestAction)
 				errs = append(errs, err)
-
-				if curRel == nil {
-					action = UninstallAction
+				status.SetStatusPhase(r.hrClient.HelmReleases(hr.Namespace), hr, apiV1.HelmReleasePhaseTestFailed)
+
+				if hr.Spec.Test.GetIgnoreFailures() {
```
seaneagan, line 377: condition needs to be negated

pkg/release/release.go
```diff
@@ -306,22 +364,32 @@ next:
 	case TestAction:
 		if hr.Spec.Test.Enable {
 			logger.Log("info", "running test", "action", TestAction)
-			if err = r.test(client, hr); err != nil {
+			status.SetStatusPhase(r.hrClient.HelmReleases(hr.Namespace), hr, apiV1.HelmReleasePhaseTesting)
```
seaneagan, line 367: why pull all these `SetStatusPhase` calls out of the `test()` method? All else equals, seems better to leave them there, to follow the pattern used in the other methods, and in case we need to call `test()` in multiple places in the future.

seaneagan (changes requested):
looks like when changing the name of the field, the opposite polarity of the field still needs to be addressed in the code. also needs a rebase.

chrisjholly:
Closed and opened #472 that has a clean commit. This PR was too messy

PR #471, opened 2020-06-25
Allow installs to proceed if no v2 and v3 release exists during migration

saada:
* When rolling out migrations across many clusters, there could be some new helm releases that were migrated. For example, if we have DEV, UAT, and PROD environments; DEV might have already migrated v2 to v3 releases, UAT might still need to do the migration, while PROD might not have had the v2 releases at all. This bugfix allows all 3 scenarios to work as expected.

### DIFF
--- a/pkg/helm/v3/converter.go
+++ b/pkg/helm/v3/converter.go
@@ -1,6 +1,8 @@
 package v3
 
 import (
+	"strings"
+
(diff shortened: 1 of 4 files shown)


hiddeco (approved):

rndstr (commented):
lgtm :ok_hand:

Issue #404, opened 2020-05-04
Deprecate status "phases"


seaneagan:
**Describe the feature**

Release status phases, which were exposed in #334 are now a [deprecated concept](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) in the kubernetes API conventions. There is [ongoing work](https://github.com/kubernetes/enhancements/pull/1624) on these conventions, but in general status conditions appear to be the path forward. These steps could be taken to move away from phases towards conditions:

1. See #395 
2. See #406
3. Reframe phases as `HelmReleaseConditionReason`s (or possibly `HelmReleaseEvent` if there is a desire to expose as kubernetes events as well) in the code base
4. remove `status.phase` (has to wait for HelmRelease v2?)

seaneagan:
3. should follow how it's done in the other toolkit components:

https://github.com/fluxcd/source-controller/blob/b078847ce8c87a412c5798fc32faabcc2d7eb58c/api/v1alpha1/condition_types.go#L55
https://github.com/fluxcd/kustomize-controller/blob/81e077811f3ed09275a8eb5a063fd8c3b61e6bc1/api/v1alpha1/condition_types.go#L56

kingdonb:
Sorry if your issue remains unresolved. The Helm Operator is in maintenance mode, we recommend everybody upgrades to Flux v2 and Helm Controller.

A new release of Helm Operator is out this week, 1.4.4.

We will continue to support Helm Operator in maintenance mode for an indefinite period of time, and eventually archive this repository.

Please be aware that Flux v2 has a vibrant and active developer community who are actively working through minor releases and delivering new features on the way to General Availability for Flux v2.

In the mean time, this repo will still be monitored, but support is basically limited to migration issues only. I will have to close many issues today without reading them all in detail because of time constraints. If your issue is very important, you are welcome to reopen it, but due to staleness of all issues at this point a new report is more likely to be in order. Please open another issue if you have unresolved problems that prevent your migration in the appropriate Flux v2 repo.

Helm Operator releases will continue as possible for a limited time, as a courtesy for those who still cannot migrate yet, but these are strongly not recommended for ongoing production use as our strict adherence to semver backward compatibility guarantees limit many dependencies and we can only upgrade them so far without breaking compatibility. So there are likely known CVEs that cannot be resolved.

We recommend upgrading to Flux v2 which is actively maintained ASAP.

[...]

PR #472, opened 2020-06-25
Adding support for optional rollback/uninstall when helm test fails

### LINKED ISSUE #454: Ability to ignore (or not) helm test failures

chrisjholly:
Fixes #454 

Linked PR #463


chrisjholly:
@hiddeco PR with clean changes from #463 with @seaneagan comments addressed.

hiddeco:
@chrisholly it appears that some of the Helm test tests are flakey, especially the `when test.enable and rollback.enable are set, releases with timed out tests are rolled back` one gave problems in #471.

docs/helmrelease-guide/tests.md
```diff
@@ -23,6 +23,18 @@ spec:
 When tests are enabled, [resource waiting](release-configuration.md#wait-for-resources-to-be-ready)
 defaults to `true` since this is likely needed for test pre-conditions to be satisfied.
 
+## Rollback release on test failure
+
+IgnoreFailures will cause a Helm release to be unistalled/rolled
```
seaneagan: this description is inverted (likely due to renaming). also would be good to mention that test failures only affect the `Released` status condition when not ignored.

test/e2e/40_tests.bats
```diff
@@ -36,7 +36,7 @@ function setup() {
   poll_until_equals 'release deploy' 'True' "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-helm-repository -o jsonpath='{.status.conditions[?(@.type==\"Tested\")].status}'"
 }
 
-@test "When test.enable is set, releases with failed tests are uninstalled" {
+@test "When test.enable is set and test.ignoreFailures is true, releases with failed tests are uninstalled" {
```
seaneagan:
```suggestion
@test "When test.enable is set and test.ignoreFailures is false, releases with failed tests are uninstalled" {
```

test/e2e/40_tests.bats
```diff
@@ -49,7 +49,7 @@ function setup() {
 }
 
 # TODO: Fail tests on install instead of upgrade once install retries can be disabled.
-@test "When tests fail, Tested and Released conditions are False" {
+@test "When tests fail and test.ignoreFailures is true, Tested and Released conditions are False" {
```
seaneagan:
```suggestion
@test "When tests fail and test.ignoreFailures is false, Tested and Released conditions are False" {
```

pkg/status/conditions.go
```diff
@@ -133,11 +133,19 @@ func ConditionsForPhase(hr *v1.HelmRelease, phase v1.HelmReleasePhase) ([]v1.Hel
...
-			Type:    v1.HelmReleaseReleased,
-			Status:  v1.ConditionFalse,
-			Message: message,
-		})
+		if hr.Spec.Test.GetIgnoreFailures() {
+			conditions = append(conditions, &v1.HelmReleaseCondition{
+				Type:    v1.HelmReleaseReleased,
+				Status:  v1.ConditionTrue,
+				Message: message,
+			})
```
seaneagan: I think this is redundant since the HelmReleasePhaseSucceeded should get set subsequently and handle this. Also when #452 is implemented we will need to be able to set this phase without it affecting the `Released` condition, so removing setting it here will get us closer to that.

test/e2e/fixtures/releases/test/fail-no-rollback.yaml
```diff
@@ -9,6 +9,7 @@ spec:
   timeout: 30
   test:
     enable: true
+    ignoreFailures: true
```
seaneagan, line 12: looks like this file was actually unused before. can we rename this file to indicate that it is about ignoring test failures e.g. `fail-ignored.yaml`?

also can we enable rollbacks in this file to show that the `ignoreFailures` and not rollbacks being disabled is what is causing the rollback not to happen. i'm wondering if that may actually be what's causing the test to pass currently as I'm not seeing any code in `release.go` to handle avoiding rollbacks / uninstalls. speaking of, can we also add a test to ensure un-installs don't happen after an install when `ignoreFailures` is set?

seaneagan (changes requested):

seaneagan:
> @ChrisHolly it appears that some of the Helm test tests are flakey, especially the `when test.enable and rollback.enable are set, releases with timed out tests are rolled back` one gave problems in #471.

It looks like the grpc helm test pod for the podinfo chart failed, but there are no logs captured for that test pod so not sure why. If we see it again I can look into addressing.

test/e2e/fixtures/releases/test/fail-no-rollback.yaml
chrisjholly: > can we also add a test to ensure un-installs don't happen after an install when ignoreFailures is set?

Have you got any suggestions how I can assert that an uninstall has not happened as the current test infrastructure only supports waiting for a condition to be become true and not lack of a condition occurring?

pkg/status/conditions.go
chrisjholly: @seaneagan I would have thought that the 'HelmReleasePhaseSucceeded' should not be called when failed tests are ignored as you would want the helm release phase to be 'TestFailed' and not 'Succeeded'. It really depends if 'TestFailed' should be treated as a terminal state of a helm release or an interim state.

test/e2e/fixtures/releases/test/fail-no-rollback.yaml
chrisjholly: > can we rename this file to indicate that it is about ignoring test failures 

done

> also can we enable rollbacks in this file to show that the ignoreFailures and not rollbacks being disabled is what is causing the rollback not to happen

done

>  i'm wondering if that may actually be what's causing the test to pass currently as I'm not seeing any code in release.go to handle avoiding rollbacks / uninstalls

code was lost in merge/commit. clearly the test did not provide enough coverage so added an extra assertion to check it is not rolled back

> can we also add a test to ensure un-installs don't happen after an install when ignoreFailures is set?

I have added an extra assertion to the existing test to validate it ignores the test failure

pkg/status/conditions.go
seaneagan: See #404, phase is a deprecated concept. The important thing is the `Reason` that gets persisted on the conditions, which with my suggestion the `Tested` condition will be left with the `TestFailed` reason and the `Released` condition will be left in the `Succeeded` reason, when `ignoreFailures` is set and tests fail.

pkg/release/release.go
```diff
@@ -364,18 +364,26 @@ next:
...
-					action = RollbackAction
+					action = AnnotateAction
+					logger.Log("info", "test failed - ignoring failures", "ignoreFailures", hr.Spec.Test.GetIgnoreFailures(), "action", action)
 				}
 				goto next
 			}
 
+			status.SetStatusPhase(r.hrClient.HelmReleases(hr.Namespace), hr, apiV1.HelmReleasePhaseTested)
 			logger.Log("info", "test succeeded", "revision", chart.revision, "action", action)
 		}
```
seaneagan, line 386: The line directly below this (L390) should be called in the case of ignored test failures, but AFAICT is not due to L383, which I believe needs to be moved to directly after L378, and then I think L386-L387 need to put in an `else` block (so it's only called when tests succeed).

test/e2e/40_tests.bats
```diff
@@ -70,7 +70,27 @@ function setup() {
...
+
+  # Assert `Tested` condition is `False`
+  poll_until_equals 'tested condition False' 'False' "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-helm-repository -o jsonpath='{.status.conditions[?(@.type==\"Tested\")].status}'"
+
+  # Assert test failures are ignored but logged
+  poll_until_true 'test failure - ignore failures' "kubectl -n $E2E_NAMESPACE logs deploy/helm-operator | grep -E \"test failed - ignoring failures\""
+
+  # Assert `Released` condition is `True`
+  poll_until_equals 'released condition True' 'True' "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-helm-repository -o jsonpath='{.status.conditions[?(@.type==\"Released\")].status}'"
+}
```
seaneagan, line 100: Can you add a check that `status.revision` is set correctly to the successful (despite ignored failed test) chart revision?

pkg/release/release.go
```diff
@@ -554,18 +562,15 @@ func (r *Release) test(client helm.Client, hr *apiV1.HelmRelease) (err error) {
 	defer func(start time.Time) {
 		ObserveReleaseAction(start, TestAction, err == nil, hr.GetTargetNamespace(), hr.GetReleaseName())
 	}(time.Now())
-	status.SetStatusPhase(r.hrClient.HelmReleases(hr.Namespace), hr, apiV1.HelmReleasePhaseTesting)
```
seaneagan: is there an advantage to moving these 3 test-related `SetStatusPhase` calls outside the `test()` method? Elsewhere we are keeping these calls inside the action method e.g. with `install()`, `upgrade()`, `rollback()`, which seems useful in case the method needs to be called in multiple places which may be a possibility with e.g. #452.

seaneagan (changes requested):

pkg/release/release.go
chrisjholly: it was required but not required anymore - so reverted back

pkg/release/release.go
chrisjholly: agreed

test/e2e/40_tests.bats
```diff
@@ -70,7 +70,24 @@ function setup() {
   [ "$output" = 'False' ]
 }
 
-@test "When test.enable and rollback.enable are set, releases with failed tests are rolled back" {
+@test "When tests fail and test.ignoreFailures is true, release has phase 'TestFailed' and Released condition is True & Tested condition is False" {
```
seaneagan: This test checks for ignored test failures on an install, it would be nice to have a test for ignored test failures on an upgrade as well. Essentially would be a copy of this test with a successful install added at the beginning.

seaneagan: Or to save on execution time could just tack the upgrade test on to the end of this test, by making some small patch to the same HelmRelease which would trigger the upgrade with ignored failed tests. Would want to ensure that the upgrade does in fact occur (e.g. checking for upgrade log message), since the expected end condition state of the upgrade will be the same as the install.

seaneagan:
```suggestion
@test "When tests fail and test.ignoreFailures is true, release has Released condition is True & Tested condition is False" {
```

chrisjholly: done

test/e2e/40_tests.bats
chrisjholly: @seaneagan do you mean:
`poll_until_equals 'revision match' "4.0.1" "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-helm-repository -o jsonpath='{.status.revision}'"`

status.revision should be the chart.version if it's not a git sync'd chart? If you mean asserting on the revision '4.0.1' I can add it but I'm not sure how much value it adds?

seaneagan: Yes I just made that suggestion since I think `SetStatusPhaseWithRevision()` was not getting called in some cases at that time, but it's probably not that useful.

test/e2e/40_tests.bats
```diff
@@ -70,7 +70,33 @@ function setup() {
...
+  poll_until_equals 'released condition True' 'True' "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-helm-repository -o jsonpath='{.status.conditions[?(@.type==\"Released\")].status}'"
+
+  # Change the number of replicas to trigger upgrade
+  kubectl patch -f "$FIXTURES_DIR/releases/test/fail-ignored.yaml" --type='json' -p='[{"op": "replace", "path": "/spec/values/replicaCount", "value":"1"}]' >&3
+
+  # Wait to make sure upgrade is successful
+  poll_until_true 'upgrade complete' "kubectl -n $E2E_NAMESPACE logs deploy/helm-operator | grep -E \"upgrade succeeded\""
+
+  # Assert `Tested` condition is `False`
+  poll_until_equals 'tested condition False' 'False' "kubectl -n $DEMO_NAMESPACE get helmrelease/podinfo-helm-repository -o jsonpath='{.status.conditions[?(@.type==\"Tested\")].status}'"
```
seaneagan, line 96: ideally would be good to check that `Released` condition is true here as well to show that the test failure was in fact ignored.

seaneagan (approved):
@chrisjholly lgtm, thanks for picking this up!

docs/helmrelease-guide/tests.md
```diff
@@ -23,6 +23,19 @@ spec:
 When tests are enabled, [resource waiting](release-configuration.md#wait-for-resources-to-be-ready)
 defaults to `true` since this is likely needed for test pre-conditions to be satisfied.
 
+## Uninstall/Rollback release on test failure
+
+The `spec.test.ignoreFailures` allows the helm release to be left in a released state if the tests fail.
+Setting ignoreFailures to false will automatically uninstall/rollback the Helm Release if any of the tests fail.
+If the tests are ignored, the `Released` condition will be left as true and `Tested` will be false.
```
hiddeco:
```suggestion
## Uninstall or rollback release on test failure

The `spec.test.ignoreFailures` allows the `HelmRelease` to be left in a released state if the tests fail.
Setting `ignoreFailures` to `false` will automatically uninstall or rollback the `HelmRelease` if any of the tests fail.
If the tests are ignored, the `Released` condition will be left as `true` and `Tested` will be `false`.
```

hiddeco (approved):
LGTM. Thanks for being patient with the requested changes @chrisjholly, and thanks a lot for the rigorous reviews @seaneagan. :tulip:

PR #478, opened 2020-07-05
Fix: longer timeout for annotator

### LINKED ISSUE #479: higher timeout for annotator
**Describe the feature**
hey, 
I am using helm operator to manage a pretty large umbrella chart and I am facing the issue where after a release upgrade helm operator gets timeout when trying to annotate all of the resources

**What would the new user story look like?**
How would the new interaction with the Helm Operator look like? E.g.
0. What are the prerequisites for this? edit this file: https://github.com/fluxcd/helm-operator/blob/master/pkg/release/annotator.go#L73
1. change the timeout to either 120 seconds or make it dynamic via a parameter passed to the helm chart

**Expected behavior**
helm operator will have less likelihood to fail upon annotating large umbrella charts. 

attached here is a PR with the wanted change:
https://github.com/fluxcd/helm-operator/pull/478

avielb:
Fixes #479

### DIFF
--- a/pkg/release/annotator.go
+++ b/pkg/release/annotator.go
@@ -70,7 +70,7 @@ func annotateResources(rel *helm.Release, resourceID resource.ID) error {
 
 		// The timeout is set to a high value as it may take some time
 		// to annotate large umbrella charts.
-		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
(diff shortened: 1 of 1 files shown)


avielb:
a little bit of attention here ?

hiddeco (approved):

PR #480, opened 2020-07-06
Add flag for disabling OpenAPI Validation for Helm 3

waseem-h:
### Requirement
Helm 3 allows us to disable OpenAPI Validation when building resources via kubernetes client and we've been using that so that we can have a helm chart that works for all kubernetes versions. An example is https://github.com/curityio/idsvr-helm/blob/master/idsvr/templates/cluster-conf.yaml#L46
It works fine with helm 2 HelmRelease but not with helm 3 without specifying the `--disable-openapi-validation` flag on the helm command. Adding it to the HelmRelease allows us to pass it to the helm 3 API so that it doesn't validate our templates


waseem-h:
CC @hiddeco @stefanprodan can any of you guys review this please?

hiddeco (approved):
Thank you @waseem-h :tulip:

PR #482, opened 2020-07-09
Add grafana helm-operator dashboard

lucasvmiguel:
It adds a new Grafana dashboard for helm-operator.

Screenshot 2020-07-09 at 12 01 22
Screenshot 2020-07-09 at 12 01 33
Screenshot 2020-07-09 at 12 01 44

### DIFF
--- a/chart/helm-operator/README.md
+++ b/chart/helm-operator/README.md
@@ -147,6 +147,7 @@ chart and their default values.
 | `readinessProbe.failureThreshold`                 | `3`                                                  | The number of times the readiness probe can failed before the container is marked as unready
 | `initContainers`                                  | `[]`                                                 | Init containers and their specs
 | `hostAliases`                                     | `{}`                                                 | Host aliases allow the modification of the hosts file (`/etc/hosts`) inside Helm Operator container. See <https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/>
+| `dashboard.enabled`                               | `false`                                              | If enabled, helm-operator will create a configmap with a dashboard in json that's going to be picked up by grafana (see [sidecar.dashboards.enabled](https://github.com/helm/charts/tree/master/stable/grafana#configuration))
 
 ## How-to
 
--- a/chart/helm-operator/README.md
+++ b/chart/helm-operator/README.md
@@ -147,7 +147,7 @@ chart and their default values.
 | `readinessProbe.failureThreshold`                 | `3`                                                  | The number of times the readiness probe can failed before the container is marked as unready
 | `initContainers`                                  | `[]`                                                 | Init containers and their specs
 | `hostAliases`                                     | `{}`                                                 | Host aliases allow the modification of the hosts file (`/etc/hosts`) inside Helm Operator container. See <https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/>
-| `dashboard.enabled`                               | `false`                                              | If enabled, helm-operator will create a configmap with a dashboard in json that's going to be picked up by grafana (see [sidecar.dashboards.enabled](https://github.com/helm/charts/tree/master/stable/grafana#configuration))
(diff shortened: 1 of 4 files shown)


lucasvmiguel:
Please, let me know if I'm missing any important metrics.

Also, it would be good if another person could test importing this dashboard in their cluster (just to check if it's working for everyone, not just for me)

chart/helm-operator/dashboards/helm-operator-dashboard.json
```diff
@@ -0,0 +1,703 @@
...
+        "label": null,
+        "name": "logs_search",
+        "options": [
+          {
+            "selected": true,
+            "text": "mysql",
+            "value": "mysql"
+          }
+        ],
+        "query": "mysql",
```
billimek: We may want to sanitize this such that the logs_search default value is empty instead of `mysql`

chart/helm-operator/dashboards/helm-operator-dashboard.json
```diff
@@ -0,0 +1,703 @@
...
+    },
+    {
+      "type": "panel",
+      "id": "logs",
+      "name": "Logs",
+      "version": ""
+    },
+    {
+      "type": "datasource",
+      "id": "loki",
```
serdardalgic, line 41: One suggestion, it would be nice to have the usage of Loki customizable. Although it's a great tool, for any other constraint there would be projects that are not using Loki. A flag that would enable/disable the logs in the dashboard would be awesome.

billimek:
Because of screen real estate, and possible superfluous information, it may be useful to reorganize the table columns in the "Helm Releases" panels, for example:
image

Specifically showing only the timestamp, release namespace, release name, and (maybe?) the helm-operator pod.  It may be redundant to show the endpoint, service, etc of the helm-operator itself:
image

The example screenshot you display shows what I suggest (and also show) but the json code seems to reflect a different set of defaults.

chart/helm-operator/dashboards/helm-operator-dashboard.json
lucasvmiguel: I've seen some dashboards in other OSS projects with Loki, it's the default way to see logs in Grafana. 

That being said, it would be nice to have this toggle, but in the end, the dashboard will still be working (with or without loki)

lucasvmiguel:
I'm not allowed to re-run the tests, but I'm almost sure that the failed test is not related to this change

chart/helm-operator/README.md
```diff
@@ -147,6 +147,7 @@ chart and their default values.
 | `readinessProbe.failureThreshold`                 | `3`                                                  | The number of times the readiness probe can failed before the container is marked as unready
 | `initContainers`                                  | `[]`                                                 | Init containers and their specs
 | `hostAliases`                                     | `{}`                                                 | Host aliases allow the modification of the hosts file (`/etc/hosts`) inside Helm Operator container. See <https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/>
+| `dashboard.enabled`                               | `false`                                              | If enabled, helm-operator will create a configmap with a dashboard in json that's going to be picked up by grafana (see [sidecar.dashboards.enabled](https://github.com/helm/charts/tree/master/stable/grafana#configuration))
```
stefanprodan:
```suggestion
| `dashboards.enabled`                               | `false`                                              | If enabled, helm-operator will create a configmap with a dashboard in json that's going to be picked up by grafana (see [sidecar.dashboards.enabled](https://github.com/helm/charts/tree/master/stable/grafana#configuration))
```

stefanprodan (approved):
LGTM

bcollard:
Many thanks @lucasvmiguel for these two Grafana dashboard contributions! (Helm Operator + Flux)

As far as I'm concerned, we are using the EFK stack in my company and I believe we are many to do so within Kubernetes.
So I had to strip out some parts related to Loki, but that's fine. 

Do you know how difficult would that be to grab logs from ElasticSearch instead?

lucasvmiguel:
I don't think you'll be able to grab logs from ES without changing the dashboard, Loki is configured with 
Loki datasource type, you'd have to use ES datasource type instead.

Reference: https://grafana.com/docs/grafana/latest/features/datasources/

gopisaba:
is there a plan to add EFK support for the grafana dashboard?

PR #492, opened 2020-07-21
Fix cast panic in operator.go

luxas:
Objects given to informed add/update/delete events are always pointers to structs, so a cast of that object to a struct (without the pointer) will always result in a panic.
In the case of this deletefunc, we've already cast the object to a pointer in `checkCustomResourceType` and returned the struct alone, so just use that in the `status.ObserveReleaseConditions` call

cc @hiddeco


hiddeco (approved):
Excellent 🎣

PR #496, opened 2020-07-22
Looping over namespaces for rbac creation as in flux chart

phisco:
In order to be able to create roles in multiple namespaces instead of a single admin clusterrole I've introduced the same logic implemented for flux's helm chart.

- added an `allowedNamespaces` parameter to the values with the list of allowed namespaces
- added logic to create role and rolebinding in all namespaces specified

In order to have full feature parity between flux and helm-operator the `--k8s-allow-namespaces` flag should be added, let me know if i can start working on that too.

### DIFF
--- a/chart/helm-operator/README.md
+++ b/chart/helm-operator/README.md
@@ -107,6 +107,7 @@ chart and their default values.
 | `logFormat`                                       | `fmt`                                                | Log format (fmt or json)
 | `logReleaseDiffs`                                 | `false`                                              | Helm Operator should log the diff when a chart release diverges (possibly insecure)
 | `allowNamespace`                                  | `None`                                               | If set, this limits the scope to a single namespace. If not specified, all namespaces will be watched
+| `allowedNamespaces`                               | `[]`                                                 | Allow helm-operator to manage resources in the specified namespaces. The namespace flux is deployed in will always be included
 | `helm.versions`                                   | `v2,v3`                                              | Helm versions supported by this operator instance, if v2 is specified then Tiller is required
 | `tillerNamespace`                                 | `kube-system`                                        | Namespace in which the Tiller server can be found
 | `tillerSidecar.enabled`                           | `false`                                              | Whether to deploy Tiller as a sidecar (and listening on `localhost` only).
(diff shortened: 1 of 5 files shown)


phisco:
Tagging @stefanprodan given that you are listed as maintainer [here](https://github.com/fluxcd/helm-operator/blob/master/chart/helm-operator/Chart.yaml#L10)

stefanprodan:
> In order to have full feature parity between flux and helm-operator the --k8s-allow-namespaces flag should be added, let me know if i can start working on that too.

A Kubernetes informer can watch the whole cluster or a single namespace, see https://github.com/fluxcd/helm-operator/blob/master/cmd/helm-operator/main.go#L261 I don't see an easy way to implement this with the current client-go.

phisco:
I meant something like [this](https://github.com/fluxcd/flux/blob/master/cmd/fluxd/main.go#L180) in flux, but I guess the two usages are different

marratj:
I think the difference here is that helm-operator uses an informer which only can watch the whole cluster or a single namespace as @stefanprodan pointed out whereas Flux itself is not using an informer at all because it "simply" looks what resources are there in a Git repo and if there would be one to sync in a namespace that is not in the allowed list it would skip by itself with its own logic.

phisco:
Yes, definitely! I was just trying to understand whether the go part was needed or not, I guess not. One thing is to have the namespace list for the rbac set up from the helm chart, another is having a list for the the operator to watch all the helmreleases in those namespaces. I'm not sure if it's possible to restrict the scope of the operator to just a set of namespaces given the current informer API, but I still think there is value in setting a fine grained rbac with rolebindings for each namespace instead of a single clusterrole binding granting admin privileges. Let me know what you think.

marratj:
Right, the first part of it is definitely worth looking into (creating Role(Binding)s for all `allowedNamespaces` instead of ClusterRoleBindings).

What actually happens if the Helm Operator itself is not configured to watch only a single namespace but has only namespace-scoped Role Bindings? Will it then still pick up the `HelmReleases` of those namespaces or does it have a problem because some cluster wide permission is missing?

phisco:
I'm currently running it with a clusterrole allowing it to handle helmreleases cluster-wide, but only rolebindings on specific namespaces, this is fine for me because this is the only helm-operator on the cluster. This can be achieved by setting `clusterRole.create: false` and a clusterrole/clusterrolebinding to allow: `[["apiextensions.k8s.io"], ["customresourcedefinitions"], [ "list", "watch"]]` and `[["helm.fluxcd.io"],["*"],["*"]]`

PR #515, opened 2020-08-13
Compare dry-run and release manifests to determine upgrades.

### LINKED ISSUE #16: Helm operator does not notice git changes to child charts
In a Helm chart, it is allowed to have relative `file://` references in your `requirements.yaml`, .e.g.:

```
❯ tree charts
charts
├── child/..
└── parent/..

❯ cat charts/parent/requirements.yaml
dependencies:
- name: child
  version: 0.1.0
  repository: file://../child
```

The Helm operator does however [only look for commits touching the configured parent chart path (e.g. `charts/parent`) on git mirror updates](https://github.com/fluxcd/flux/blob/daad1ff216b17db4560d7a154e59cc5b197fd0f2/integrations/helm/chartsync/chartsync.go#L219-L221), which causes changes to child charts to go unnoticed. Subsequent dry-runs do not pick up the change either, as the clone for a `HelmRelease` [is only replaced when a change is detected](https://github.com/fluxcd/flux/blob/daad1ff216b17db4560d7a154e59cc5b197fd0f2/integrations/helm/chartsync/chartsync.go#L240-L242).

The easiest and neatest fix would probably be to always replace the clone of a chart, but to only schedule a release on changes to the parent chart. This would ensure changes to the parent chart result in a priority release, while changes to child charts are still picked up after a while by the sync / mutation prevention mechanism.

CarolineTeicher:
Fixes #16 

If applied, this commit will allow helm-operator, on `dryRunCompare`, to notice changes made to child charts. This will not permit spurious upgrades.

### DIFF
--- a/pkg/helm/utils.go
+++ b/pkg/helm/utils.go
@@ -5,5 +5,5 @@ import (
 )
 
 func Diff(j *Release, k *Release) string {
-	return cmp.Diff(j.Values, k.Values) + cmp.Diff(j.Chart, k.Chart)
+	return cmp.Diff(j.Values, k.Values) + cmp.Diff(j.Chart, k.Chart) + cmp.Diff(j.Manifest, k.Manifest)
 }


CarolineTeicher:
Hi @stefanprodan - is there an update on this? Thanks!

stefanprodan:
@CarolineTeicher I think a go template function that generates timestamps or random strings will cause spurious upgrades.

amit-handda:
so, IIUC, helm-operator wont catch a diff between the generated manifest. but only if the chart's metadata or template contents are updated explicitly ?

In our case, I am generating configmap via template. The configmap yaml can update as per the files matched via `glob` pattern. And I have a corresponding `shasum` of the configmap attached to the deployment pod, so that the pod ll redeploy. 

```apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    openpolicyagent.org/policy: rego
{{ include "opa.labels.standard" . | indent 4 }}
  name: opa-policies-configmap
  namespace: {{ .Release.Namespace }}
data:
{{ $root := . }}
{{ range $path, $bytes := .Files.Glob "*/common/**.rego" }}
  {{ base $path }}: |-
{{ $root.Files.Get $path | indent 4 }}
{{ end }}
``` 

Thanks,

hiddeco:
Thank you for submitting this pull request! :tulip: 

As charts may contain generated values, we can not accept this change as those will result in drift for a set of users.

In case this is still causing problems for you, I would advice you to look into the [helm-controller](https://github.com/fluxcd/helm-controller). This controller has a different design and is better capable of working with templates like the one @amit-handda shared here.

PR #519, opened 2020-08-20
[Helm Test] Continuous helm test parameter

gmtstephane:
Hello,
This PR is about adding a continuous Helm test feature to trigger tests on a regular basis.

1. Added Continuous definitions in crds and go structures
```
continuous:
    description: Set Continuous to true will run test continuously.
    type: boolean
```

2. Trigger test if test and continuous parameters are enabled
```
 if hr.Spec.Test.Enable && hr.Spec.Test.Continuous == true {
	logger.Log("info", "Triggering continuous test", "phase", action)
	action = TestAction
	goto next
}
```

### DIFF
--- a/chart/helm-operator/crds/helmrelease.yaml
+++ b/chart/helm-operator/crds/helmrelease.yaml
@@ -206,6 +206,9 @@ spec:
                     to delete test pods between each test run initiated by the Helm
                     Operator.
                   type: boolean
+                continuous:
+                  description: Set Continuous to true will run test continuously.
+                  type: boolean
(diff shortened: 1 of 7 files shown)


hiddeco:
Thank you for submitting this pull request, and sorry it took so long to get back to you :sunflower: 

As the Helm Operator is in maintenance mode, adding this new feature would be too much. If you are still interested in such a feature, I would advise you to open an issue for the [helm-controller](https://github.com/fluxcd/helm-controller), so we can discuss details there.

PR #522, opened 2020-08-26
Add more Helm chart dashboard options

tux-00:
- Rename `flux-dashboard.json` into `flux.json` to avoid naming like flux-dashboards-flux-dashboard
- Add `nameprefix` into dashboards values to allow better naming possibilities
- Add `namespace` into dashboards values to allow users to put dashbards in a different namespace

See https://github.com/fluxcd/flux/pull/3183


tux-00:
Strange error, could someone restart the CI ?

uberspot (approved):
Sorry for pinging @hiddeco but would it be possible to get this merged? 
This would allow changing the name of the dashboard configmap per namespace and would make loading these dashboards automatically doable with something like grafana sidecars https://github.com/helm/charts/tree/master/stable/grafana#configuration when we have multiple helm-operators deployed in a cluster. 
 
Right now if you load all these configmaps with the sidecar they overwrite each other because the name is the same for all of them. 

Specifically from grafana sidecar:
```
t=2020-10-02T11:02:32+0000 lvl=eror msg="the same 'uid' is used more than once" logger=provisioning.dashboard type=file name=sidecarProvider uid=...redacted... provider=sidecarProvider

```

hiddeco (approved):

uberspot:
Nice! :) thank you! Checks passed too.

PR #529, opened 2020-09-09
Update chart-sources.md

neil-greenwood:
Add line-breaks so bullet list shows correctly on https://docs.fluxcd.io/projects/helm-operator/en/stable/helmrelease-guide/chart-sources/


hiddeco (approved):
Thank you @neil-greenwood :tulip:

Issue #457, opened 2020-06-09
Constantly creating new releases for charts even when no changes


brpaz:
**Describe the bug**

Hello.

I am starting experimenting with Flux and the Helm Operator on a new Cluster and everything went fine until I deployed cert-manager Helm chart.

Each time the sync runs, the helm operator tries to do an update and create a new release even without any change in the chart. 

This is causing some instability in the Network of my cluster. (maybe to do excessive load in the API server resulted from the constant updates.

What is strange is that if I run: ```kubectl -n  cert-manager get helmreleases.helm.fluxcd.io``` the latest update date is the initial deploy. Still a new secret with the helm release information is being created every time and the pods restarted.

**To Reproduce**

Just a basic helm release manifest:

```
---
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  name: cert-manager
  namespace: cert-manager
spec:
  releaseName: cert-manager
  chart:
    repository: https://charts.jetstack.io
    name: cert-manager
    version: 0.15.1
  values:
    installCRDs: true
    global:
      leaderElection:
        namespace: cert-manager
    ingressShim:
      defaultIssuerName: letsencrypt-prod
      defaultIssuerKind: ClusterIssuer
    prometheus:
      enabled: false
```

**Expected behavior**

Cert-manager should only be deployed when the is any change.

**Logs**

Here is the log output:

```
´s=2020-06-09T11:27:43.485024459Z caller=helm.go:69 component=helm version=v3 info="checking 31 resources for changes" targetNamespace=cert-manager release=cert-manager
ts=2020-06-09T11:27:43.511728151Z caller=helm.go:69 component=helm version=v3 info="Looks like there are no changes for ServiceAccount \"cert-manager\"" targetNamespace=cert-manager release=cert-manager
ts=2020-06-09T11:27:43.52932945Z caller=helm.go:69 component=helm version=v3 info="Looks like there are no changes for ServiceAccount \"cert-manager-webhook\"" targetNamespace=cert-manager release=cert-manager
ts=2020-06-09T11:27:43.593238042Z caller=helm.go:69 component=helm version=v3 info="Looks like there are no changes for CustomResourceDefinition \"certificaterequests.cert-manager.io\"" targetNamespace=cert-manager release=cert-manager
[...]

PR #533, opened 2020-09-16
Address constant release creation

mnaser:
This pull request addresses the issue at #457 or at least one flavor of it that I have personally identified to cause loops to create releases non-stop.

### DIFF
--- a/pkg/helm/utils.go
+++ b/pkg/helm/utils.go
@@ -1,9 +1,20 @@
 package helm
 
 import (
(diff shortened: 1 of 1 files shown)


hiddeco:
I think a better place to address this conversion would be here: https://github.com/fluxcd/helm-operator/blob/master/pkg/helm/v3/release.go#L70, as this would solve it throughout the application.

hiddeco (approved):

PR #549, opened 2020-10-05
doc fix: Malformed list in Chart sources -> Git repositories

mattjw:
The "_The definition of the listed keys is as follows_" on the _Chart sources_ is malformed when its rendered. As per [here](https://docs.fluxcd.io/projects/helm-operator/en/latest/helmrelease-guide/chart-sources/#git-repositories) and this screen grab:

helm screen grab

The list should display as three items.


hiddeco (approved):
LGTM, thank you @mattjw :100:

PR #551, opened 2020-10-07
fix: make passing gitconfig as a value optional

Carles-Figuerola:
# Description

Adding a boolean option to not create the kubernetes secret when `git.config.enable` is set to true. This allows gitconfigs with actual secret information to be stored in kubernetes through a different pipeline rather than have actual secrets stored in the values section of the helm release (valid for autmated deploys where the deployment metadata is stored somewhere public)

As an example, for simplifying git auth in our infrastructure, we use a gitconfig like this to transparently authenticate all git operations:

```
url "https://<USERNAME>:<PASSWORD>@git.mydomain.com/"]
        insteadOf = https://git.mydomain.com/
        insteadOf = ssh://git@git.mydomain.com/
        insteadOf = git@git.mydomain.com:
```


hiddeco (approved):
LGTM, thank you @Carles-Figuerola :sunflower:

PR #564, opened 2020-11-10
Add `terminationGracePeriodSeconds` value to Helm Operator Chart

coultenholt:
This is necessary when timeout may be set to a higher value to help ensure that Helm Operator isnt killed before its done waiting for an operation to finish. Also correct spelling.

As outlined here: https://github.com/fluxcd/helm-operator/blob/master/docs/helmrelease-guide/release-configuration.md#configuring-the-timeout


arobert-delfingen:
Given that the default timeout when upgrading a chart is 300s, why not also set the default terminationGracePeriod to 300s ?

coultenholt:
> Given that the default timeout when upgrading a chart is 300s, why not also set the default terminationGracePeriod to 300s ?

Good point. Updated!

hiddeco (approved):
LGTM :+1:

Can you please rebase your PR, smash the commits, and sign-off your commit? https://github.com/fluxcd/helm-operator/pull/564/checks tells you how to do the latter.

PR #571, opened 2020-12-01
Remove the query string from the CleanURL

smarthall:
We are using Azure Blob storage as a private Helm repository. As described [here](https://cwienczek.com/2017/10/setting-up-secure-helm-chart-repository-on-azure-blob-storage/). This breaks in helm-operator because the SAS token makes the cache directory too long and it fails with the following message:

```
synchronization of release 'demo' in namespace 'default' failed: failed to prepare chart for release: chart unavailable: mkdir /tmp/v3/[LONG FILE NAME]: file name too long
```

This PR strips the query string from the repo url that is used for caching so that:
1. If the authentication token changes the same cache can still be used.
2. The directory name is shorter

### DIFF
--- a/pkg/apis/helm.fluxcd.io/v1/types_helmrelease.go
+++ b/pkg/apis/helm.fluxcd.io/v1/types_helmrelease.go
@@ -3,6 +3,7 @@ package v1
 import (
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


hiddeco (approved):
LGTM, thank you very much! This should be fine to be included in the next patch release :1st_place_medal: 

Before I merge, can you please rebase your PR?

smarthall:
You are welcome, thanks for merging it.

I've rebased onto the latest master.

PR #579, opened 2021-01-11
Tighten pod security policy and allow custom role rules

langecode:
Thank you very much for a great product. I know the current Helm Operator version is in maintenance mode and it may not make sense to do this change. However, as we have done it I thought we should give you the opportunity to include it upstream should you so decide.

The PR tightens the pod security policy for Helm Operator a bit and also adds security context to the pod specifications to make sure the most correct pod security policy is chosen (see [Policy Order](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#policy-order)). Furthermore the PR will make the set of rules configured for the namespace or cluster wide role configurable making it possible to tighten the permissions applied to Flux. The default is complete access as it is in the current chart version.

Note the cluster wide objects has been prefixed with the namespace as Helm v3 works with release names as "namespaced" hence if Flux were to be deployed into multiple namespaces using the same release name there could be a clash in the naming of non-namespaced objects.

### DIFF
--- a/chart/helm-operator/templates/_helpers.tpl
+++ b/chart/helm-operator/templates/_helpers.tpl
@@ -47,11 +47,34 @@ Create the name of the cluster role to use.
 */}}
 {{- define "helm-operator.clusterRoleName" -}}
 {{- if .Values.clusterRole.create -}}
-    {{ default (include "helm-operator.fullname" .) .Values.clusterRole.name }}
+{{- $name := printf "%s-%s" .Release.Namespace (include "helm-operator.fullname" .) -}}
+{{ default $name .Values.clusterRole.name }}
 {{- else -}}
-    {{ default "default" .Values.clusterRole.name }}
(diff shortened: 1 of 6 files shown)


hiddeco:
Thank you (still) for submitting this pull request! :1st_place_medal: 

As you already mention yourself, we are in maintenance mode. Due to this, and the difficulties that at times come with PSPs, I find it too risky to merge this.

I'll keep the PR open so people that are looking for a solution like this can easily find it and copy the adjustments you made.

langecode:
Yeah, I perfectly understand. The current version of the chart will run on a cluster having PSP enforced and given the maintenance state and that the psp is never going to go out of beta it makes sense.

kingdonb:
Thanks for contributing here! We are not going to make any changes like this here, Flux v1 and Helm Operator are in maintenance mode

Closing, stale.

PR #599, opened 2021-04-07
Update CRD version

phillebaba:
This change will update the CRD version from v1beta1 to v1 in preparation for k8s v1.22. I have documented the tests that I have run to determine the different behaviors of helm-operator in different scenarios. Hopefully it is easy to follow along and understand what has been tested.


hack/update/generate-crds.sh
```diff
@@ -51,7 +51,9 @@ echo "Generating OpenAPI v3 schemas for chart CRDs"
 bin/controller-gen \
   schemapatch:manifests="${CRD_DIR}" \
   output:dir="${CRD_DIR}" \
+  crd:crdVersions=v1 \
   paths=./pkg/apis/...
+mv ${CRD_DIR}/helm.fluxcd.io_helmreleases.yaml ${CRD_DIR}/helmrelease.yaml
```
phillebaba: This is a weird behavior that the file will be different when the crd version is set. I do not know why this happens and I could not figure out how to fix it.

hiddeco: I think an option may be to use `output:stdout` (instead of `output:dir=...`) and then write it directly to the file, instead of moving one around.

deploy/crds.yaml
```diff
@@ -1,24 +1,27 @@
+---
```
stefanprodan, line 1: Is this some code gen error? We have an empty yaml doc then `v1beta1` CRD, shouldn't it be `v1`?

hack/update/generate-crds.sh
phillebaba: Thanks for the tip, makes it a lot cleaner. I did however have to trim the first two lines of the generated CRD as the new line and `---`caused other issues.

deploy/crds.yaml
phillebaba: Good catch this issue was caused by some extra new lines in the crd generation. Should be fixed now by removing them.

phillebaba:
I do not have access to the Circle CI tests so I cant see why the tests are failing. Running `make test`locally works so that may not be the issue.

hiddeco:
The whole end-to-end test suite fails due to the operator not installing correctly.

phillebaba:
Ah missed that, looks like I should be able to run e2e tests locally to debug this?

hiddeco:
Should work with `make e2e`.

phillebaba:
@hiddeco so the E2E tests works now. I was looking at the totally wrong location, the issue was never in the tests but rather that values were not being parsed properly. Had to understand the tests before I could understand that. Can you have a look now?

kingdonb:
I incorporated this PR in #623 and will rebase it now manually to match that, so it closes out when that one is merged.

kingdonb:
It rebased cleanly 👍

hiddeco (approved):
Thank you @phillebaba :beers:, and thanks @kingdonb for driving the release :train:

PR #617, opened 2021-06-18
improve log printing

wujiangfa-xlauncher:
I think it is necessary to increase the log so that we can know what happened through the log


kingdonb:
Hello! Thanks for contributing. Logging changes are considered breaking changes, and as Helm Operator is in maintenance mode, it will not have any major upgrades going forward, it will not be possible to merge this. Please upgrade to Flux v2!

PR #618, opened 2021-06-21
Update RBAC version

kingdonb:
Piggybacking off #599, this PR removes the last vestiges of `rbac.authorization.k8s.io/v1beta1` in favor of `rbac.authorization.k8s.io/v1` that has been available since Kubernetes 1.8. This is also required for K8s 1.22.x next release.

I gave this some testing with a chartmuseum and the test suite that #599 provided, and I was able to install it on Kubernetes 1.22.0-alpha.3 which I've been working on to be sure we catch all the deprecations.

There is a PR at kingdonb/helm-operator#1 where I am noodling around with the `crd-test` chart, to prove to my satisfaction that Helm Operator can install charts based on the new CRD version (which I will open against my own fork, for now; I'm not sure how to incorporate them for merging just yet.)


kingdonb:
This PR is already in 1.4.0-rc1 👍

hiddeco (approved):
:100:

PR #620, opened 2021-06-23
Release 1.3.0-rc2

kingdonb:
The notes in Chart.yaml for Helm Operator 1.3.0-rc1 had incorrectly signaled that Kubernetes 1.16 was already required now for this release. In fact this requirement will not take effect until Helm Operator 1.4.0, the next MINOR release; the point of 1.3.0 is primarily to get a clean dependency scan for CVEs using Snyk, and other updates that haven't been published though they were merged throughout the year.

Since we say we support Kubernetes back to version 1.11, I've also decided to adjust the testing again in order to see if we can still use kind to **test** that far back. It should work. The build pipeline at CircleCI was a bit noisy for 1.3.0-rc1 due to apparently some intermittent network issues but everything did come out OK, all e2e tests were passing, and an image was pushed. 🎉  I haven't found any issues with it.

There may be no changes in `rc2` or even need for second candidate. There are two candidate PRs simply so I can run the CI systems again, with a different setting for KUBE_VERSION.

I'll be publishing `1.4.0-rc1` next, then Flux `1.23.0-rc1`, all before coming back to certify each release of Helm Operator and Flux. The planned release date of all three is Friday!


kingdonb:
This is identical to #622 except for the version bump, in cbc26bd0

(Closing, as there was no candidate build)

PR #622, opened 2021-06-25
Release 1.3.0

kingdonb:
The changelog has been added in #620 - this is the last update prior to CRD v1 and the next version requires Kubernetes 1.16+ or greater.

The updates were tested in parallel with #621 (for Release 1.4.0), these builds passed all release validation checks.

PR #621, opened 2021-06-23
Release 1.4.0-rc1

kingdonb:
This release has breaking changes to support Kubernetes 1.22.0. See the CHANGELOG.

This release candidate is based on 1.3.0, which is also a release candidate for now, and this will obviously need to be reconciled with #620 in the `release/1.4.x` branch, before it can ultimately be merged back to master.

PR #623, opened 2021-06-25
Release 1.4.0

kingdonb:
This release updates `CustomResourceDefinition` to `v1`, requiring an update to Kubernetes v1.16 or greater. This allows users to upgrade to K8s 1.22.0 and for the foreseeable future versions.

It is a breaking change, and #622 is the alternative path for users who are not ready to upgrade to 1.16. An admonishment was added in #621 to ensure users who consult the changelog will know about this requirement (the Chart.yaml should enforce the requirement just as well, but it was placed in the changelog for better visibility.)


kingdonb:
I think I could have done a bit better with the git branching here, I will handle the merges manually to ensure the result is correct. I'm putting together two more PRs to release charts that correspond to these releases.

I decided not to do a pre-release/devel chart. I tested with the method described in #599 instead, with a local install of chartmuseum placed alongside of the test deployments, in the target/test cluster.

hiddeco (approved):

PR #636, opened 2021-11-01
replace deprecated io/ioutil in packages

DiptoChakrabarty:
This PR is to replace the io/ioutil package which has been deprecated , the replacement has been done on all the various packages used in the helm operator

### DIFF
--- a/internal/lockedfile/lockedfile.go
+++ b/internal/lockedfile/lockedfile.go
@@ -9,7 +9,6 @@ package lockedfile
 import (
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"runtime"
 )
@@ -103,7 +102,7 @@ func Read(name string) ([]byte, error) {
 	}
 	defer f.Close()
 
-	return ioutil.ReadAll(f)
+	return io.ReadAll(f)
 }
 
 // Write opens the named file (creating it with the given permissions if needed),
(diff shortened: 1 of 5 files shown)


kingdonb:
Thanks for contributing this! It looks like some failures were introduced at compile time.

Flux v1 and Helm Operator are formally superseded by [Flux v2](/fluxcd/flux2) and its component part, the Helm Controller, which are projects that are actively developed together, also known as the GitOps Toolkit. 

I wonder also, if you've seen the upgrade advisory posted in the docs here: https://fluxcd.io/legacy/helm-operator/

or the Migration and Support Timetable here:

> June 30, 2021 | Superseded
> Flux 1 Helm Operator code freeze – no further updates except CVEs
> https://fluxcd.io/docs/migration/timetable/

This might not be something which we can merge even when fixed, according to our migration and support policy.

Can you say a bit more about what upgrades it may enable to replace the deprecated `ioutil` package? I am not especially familiar with this issue, I would like to understand more and know if there's a way this might be framed as addressing a CVE.

Is this package vulnerable, or is it removed in future Go versions? (Is it still present in Go 1.17? At a glance, it appears to be.)

Unless it prevents some upgrade which would mitigate a CVE, I tend to think we should leave the code alone as it is, in accordance with the published policy. This may not be the case in the future if for example we learn about a CVE that demands a change, (but by then we may have ended support for the Helm Operator. It will not be supported forever.)

kingdonb:
I see you've been rooting out and addressing these issues in various repos across the fluxcd org, thanks for that.

I think I can assume then, that you are not a Helm Operator user and do not need any upgrade advice 😉

kingdonb:
Closing, stale.

PR #640, opened 2021-11-26
Housekeeping: merge release/1.4.x back into master 

kingdonb:


kingdonb:
The image was pushed successfully, in spite of the marked failure. 👍

hiddeco (approved):

kingdonb:
Looks like we are actually going to need the build to pass again in order to finally merge the housekeeping PR.

There is a flaky test, and there was a failure because of "tag already uploaded" as well. I've only seen the flaky test fail once or twice, having run the helm-operator e2e suite at least 5 or 6 times now. I think this execution will pass, then I can merge it and close this one out. @hiddeco please verify whenever you can that the published `fluxcd/helm-operator:1.4.1` passes all of the same CVE checks that were done pre-release.

There was an error while tagging the image, and I'm about 98% certain that what was pushed matches what is tagged here:

* https://github.com/fluxcd/helm-operator/commit/f6fd8f3fbeba259fd00d21c17545835a7db98a7d

however there is still that 2% bit of doubt due to the series of errors that occurred, (not all of which were due to the flaky test.) I would appreciate it if you could scan the binary once again to confirm. 👍

kingdonb:
I had a poke around inside of the image that was published and I'm satisfied, it looks like the failure scenario that I was worried about did not occur. If the tests should fail I will happily run them again, but as they take close to an hour to finish, if this scenario repeats in the future, I wonder if there's any way that we can override the flaky test? I think some tests are already skipped (perhaps also due to flakiness.)

Of course I'm sure this issue has been solved in Helm Controller 🙃

PR #648, opened 2022-03-29
Mitigate GHSA advisories

kingdonb:
It looks like certain Flux dependency upgrades will be easier if we just go ahead and bring Helm Operator up to date at the same time.

I do not know for sure when I'll be doing another Helm Operator release, but this seems to be a decent start at the most glaring issues.


kingdonb:
CI passed, so if it would make sense, we can merge some changes to resolve some CVEs in those dependencies, and I'll cut a new release in the 1.4.x series, the `release/1.4.3` branch is already open.

PR #652, opened 2022-05-18
Backport helm-controller helm release suspend via annotation

cbalan:
Added helm.fluxcd.io/suspend annotation to replicate helm-controller release suspend behaviour.

Required to enable migration to helm-controller without disabling helm-operator or making RBAC changes.


kingdonb:
Thanks for your contribution!

> Required to enable migration to helm-controller without disabling helm-operator or making RBAC changes.

Could you say more about this? Are your environments running helm-controller and helm-operator on the same clusters mixed? (Have you been able to migrate fully, or is there something in particular blocking you from turning off Helm Operator?)

Sorry for the delay in response, I'm just going through open PRs now and intended to close them all, but yours looks like it was intended to address a specific migration issue that you had I didn't understand, so I wanted to follow up about that.

kingdonb:
I think I understand you needed to be able to turn off HelmReleases in the Helm Operator one-by-one, and this implements a feature 👍 this wasn't supported before

Good work, I'm assuming you did not wait for us to merge this. I am going to close it as we cannot support new features in Helm Operator, but I appreciate that anyway! Thanks for contributing to Flux 🎉

PR #661, opened 2022-08-30
Update Flux dependency to 1.25.4

kingdonb:
Looking to see if we can easily land this update here:

* https://github.com/fluxcd/flux/pull/3636

(Let's run the CI and see how this change goes, something tells me there's a reason that Flux here is still pinned to 1.18...)


kingdonb:
This passed the tests, so I'll try it out locally some time this week, and then we can publish a version of Helm Operator with the same stance on default telemetry. 👍 

There should be no code changes required to get the new checkpoint code from fluxcd/flux/pkg/checkpoint, just to point at the newer version of Flux as a dependency.

I have not tested this at all outside of the CI scaffolding.

stefanprodan (approved):

pjbgf:
@kingdonb we need a replace to update `github.com/containerd/containerd` to `v1.5.13` to fix CVE-2022-31030.
And another for `github.com/prometheus/client_golang` to `v1.12.2` for CVE-2022-21698.

pjbgf (approved):
LGTM

